PR #29, opened 2021-01-28
WIP PR

Abuelodelanada:


metadata.yaml
```diff
@@ -20,4 +20,4 @@ resources:
     type: oci-image
     description: upstream docker image for MySQL
     auto-fetch: true
-    upstream-source: 'mysql:8.0.22'
+    upstream-source: "ubuntu/mysql:latest"
```
balbirthomas, line 23: It may be worth check with @jlounder and @jameinel what the recommended practice is - If we should be using `latest` or specific version.

src/charm.py
```diff
@@ -14,74 +16,72 @@
...
         self.framework.observe(self.on.config_changed, self._on_config_changed)
 
     @property
-    def cluster_size(self) -> int:
-        """Return the size of the cluster."""
-        rel = self.model.get_relation(PEER)
-        return len(rel.units) + 1 if rel is not None else 1
+    def hostname(self) -> str:
+        """Return the Kubernetes hostname."""
+        unit_number = self._get_unit_number_from_unit_name(self.unit.name)
```
balbirthomas, line 44: If I am not mistaken you may be able to get the unit number from something like `self.unit.id` .

src/charm.py
```diff
@@ -14,74 +16,72 @@
...
-        query = """
-                CREATE USER 'idcAdmin'@'%' IDENTIFIED BY '{}';
-                GRANT ALL ON *.* TO 'idcAdmin'@'%' WITH GRANT OPTION";
-                """.format(self.model.config['idcAdmin-password'])
 
-        cnx = self._get_sql_connection_for_host(hostname)
-        cur = cnx.cursor()
-        cur.execute(query)
-        cnx.close()
+        self.unit.status = ActiveStatus("MySQL is up and running!")
```
balbirthomas, line 84: I believe that it is recommended to not set a message for `ActiveStatus` except in rare cases.

balbirthomas (approved):

src/charm.py
Abuelodelanada: AFAIK there is no `id` property in [ops.model.Unit](https://ops.readthedocs.io/en/latest/index.html?#ops.model.Unit) or should I check somewhere else?

src/charm.py
Abuelodelanada: ack

metadata.yaml
Abuelodelanada: The available versions are these: https://hub.docker.com/r/ubuntu/mysql/tags?page=1&ordering=last_updated

balbirthomas: I meant fixing the version to a specific one rather than just picking the latest.

src/charm.py
balbirthomas: I could be wrong. I will dig when I get some time. If I find it will let you know

tests/test_charm.py
```diff
@@ -8,8 +8,73 @@
...
-    def setUp(self):
-        harness = Harness(MySQLCharm)
-        self.addCleanup(harness.cleanup)
-        harness.begin()
+    def test_hostname_for_leader_unit(self):
+        self.harness.set_leader(True)
+        expected_host = "mysql-0.mysql-endpoints"
+        self.assertEqual(self.harness.charm.hostname, expected_host)
+
+    def test_env_config(self):
```
balbirthomas, line 22: Here is my humble opinion, take it for whatever it is worth. Seems to me you are testing implementation rather than functionality of your Charm in this and a few other unit tests. Your charm has not implemented the code that makes environment variables work. This code is part of the Operator Framework/Juju/Testing Harness. Hence I see no reason why your charm should be testing that environment variables work as expected. On the other hand if one specific environment variable changes the behaviour of your charm in a specific way, then yes a unit test to check that the behaviour of your charm changes in response to that environment variable will be useful.

config.yaml
```diff
@@ -8,15 +8,17 @@ options:
...
     default: 3306
     description: Port used by MySQL
-  root-password:
+  MYSQL_ROOT_PASSWORD:
     type: string
     default: ''
-    descriptions: Password used by root user
-  idcAdmin-password:
+    descriptions: Set the password for the root user.
+  MYSQL_USER:
```
dstathis, line 15: My understanding is that setup like this should be done via a relation.

@jlounder Could you comment?

src/charm.py
```diff
@@ -98,68 +99,80 @@ def _configure_pod(self):
...
+            "password": self._stored.MYSQL_ROOT_PASSWORD,
+            "host": self.hostname,
+            "port": self.model.config["port"],
         }
         return mysql.connector.connect(**config)
 
-    def _get_unit_hostname(self, _id: int) -> str:
+    def _get_hostname_from_unit(self, _id: int) -> str:
         """Construct a DNS name for a MySQL unit."""
         return "{0}-{1}.{0}-endpoints".format(self.model.app.name, _id)
```
dstathis, line 165: I don't think this will work. The unit number does not always start at 0 but the hostnames do. For example, in my currently deployed charm, the unit name is cassandra/44 while the hostname is cassandra-0.cassandra-endpoints.lma.svc.cluster.local.

src/charm.py
```diff
@@ -98,68 +99,80 @@ def _configure_pod(self):
...
 
+    def _get_unit_number_from_unit_name(self, unit_name: str) -> int:
+        """Returns the unit number based in the unit name."""
+        UNIT_RE = re.compile(".+(?P<unit>[0-9]+)")
+        match = UNIT_RE.match(unit_name)
```
dstathis, line 170: This doesn't need to be changed as it works. It, however, is probably more efficient (and readable) to use unit_name.split('/')[1]

src/charm.py
dstathis: Once we are able to run code on in the application pod, you will be able to get this by looking in /etc/hosts

Abuelodelanada: > I don't think this will work. The unit number does not always start at 0 but the hostnames do. For example, in my currently deployed charm, the unit name is cassandra/44 while the hostname is cassandra-0.cassandra-endpoints.lma.svc.cluster.local.

Oh! I thought that **always** the unit number was the same as the number in the hostname.
Maybe I did not realise that because in every test I destroyed the model, and add it again

Abuelodelanada: I think I can get the internal address with this:

```python
    @property
    def bind_address(self) -> str:
        """The ingress-address of the MySQL server"""
        return str(self.model.get_binding(PEER).network.bind_address)
```

It seems it's working...

Abuelodelanada: Fixed

src/charm.py
Abuelodelanada: As we use bind_address, the method it's not longer neccesary

README.md
```diff
@@ -1,12 +1,85 @@
...
+Create a Juju model for your operators, say "mysql"
+
+```bash
+$ juju add-model mysql
+```
+
+Deploy a single unit of MySQL using its default configuration
+
+```bash
+$ juju deploy ./mysql.charm --resource mysql-image=ubuntu/mysql:lates
```
balbirthomas, line 46: Typo ? (`lates` or `latest` ?)

README.md
```diff
@@ -1,12 +1,85 @@
...
+```
+
+As in this example we did not especify the password for the root user, the charm will generate one, and you can get it watching the juju debug-log:
```
balbirthomas, line 49: Spelling "specify" not "especify".

README.md
```diff
@@ -1,12 +1,85 @@
...
+```bash
+$ juju debug-log -l
+
+...
+
+application-mysql: 13:45:16 WARNING unit.mysql/0.juju-log The randomly generated MYSQL_ROOT_PASSWORD is: T4NgXUeXYlJBBD37icuM
+application-mysql: 13:45:16 WARNING unit.mysql/0.juju-log Please change it as soon as possible!
+```
+
+If required, yo can remove the deployed completely:
```
balbirthomas, line 61: grammar: "deployment" not "deployed". Also spelling "you" not "yo"

README.md
```diff
@@ -1,12 +1,85 @@
...
+
+```bash
+$ juju destroy-model -y mysql --no-wait --force --destroy-storage
+```
+
+Note the `--destroy-storage` will delete any data stored by MySQL in its persistent store.
+
+
+### Parameters
+
```
balbirthomas, line 71: Parameters ? You mean `config` ?

src/charm.py
```diff
@@ -14,76 +14,63 @@
...
 
     @property
-    def cluster_size(self) -> int:
-        """Return the size of the cluster."""
-        rel = self.model.get_relation(PEER)
-        return len(rel.units) + 1 if rel is not None else 1
+    def mysql_root_password(self) -> str:
+        if "MYSQL_ROOT_PASSWORD" not in self._stored.mysql_setup:
+            password = "".join(
+                random.choice(ascii_letters + digits) for x in range(20)
```
balbirthomas, line 38: Should the password generation only be done by the leader ? If so, would it be a good idea to have a guard here to check that this method is being invoked by the leader ?

src/charm.py
```diff
@@ -98,68 +85,70 @@ def _configure_pod(self):
         self.model.pod.set_spec(spec)
         self.unit.status = ActiveStatus()
 
-    def _build_pod_spec(self):
+    def _build_pod_spec(self) -> dict:
+        """This method builds the pod_spec"""
+        if not self.unit.is_leader():
```
balbirthomas, line 90: This is good. So pod spec is only set by leader. And since `self.env_config` is only used in `_build_pod_spec`, only the leader will set the password in the environment of all the charm units. Until here everything is fine. However Juju can change the leader at any time. Or worse your leader unit may die and a new leader may get elected. So how does the new leader know that the old leader had already set a password ? I think your charm may also need to detect and handle `leader_elected` and `leader_settings_changed` events. In handling these events the old leader may need to tell the new leader the password. May be a good idea to get @jlounder opinion on this too.

balbirthomas (approved):
Please see my long comment on the possibility of leader being changed. If @jlounder thinks this is indeed and issue you may need to add further event handlers to deal with them.

README.md
Abuelodelanada: Typo!
Fixed!

README.md
Abuelodelanada: Typo! Fixed!

README.md
Abuelodelanada: Fixed!

README.md
Abuelodelanada: Fixed!

src/charm.py
Abuelodelanada: You are right, it is a possible point of failure. Fixed!

src/charm.py
Abuelodelanada: > Or worse your leader unit may die and a new leader may get elected

I had 2 units and simulated this by running: `microk8s.kubectl delete pod mysql-0 -n mysql` 
With this I deleted the leader unit. After the unit was deleted, juju automatically created a new leader unit and this unit was created using the `pod_spec` created when I deployed the application with 2 units for the first time, so the new leader unit uses the `MYSQL_ROOT_PASSWORD` generated when the application was deployed.

Certainly I need to write a test for this.

config.yaml
Abuelodelanada: These are parameters you can send to the [image we are using](https://hub.docker.com/r/ubuntu/mysql).
In the specific case of `MYSQL_ROOT_PASSWORD` if you don't set one, the charm will generate one, and will be sent vía relation data to the others units

dstathis (approved):

PR #35, opened 2021-04-05
MySQL lib first approach

Abuelodelanada:


Abuelodelanada:
With the last added test I get 80%

```bash
$ coverage report -m src/*.py
Name                   Stmts   Miss  Cover   Missing
----------------------------------------------------
src/charm.py             126     16    87%   46-50, 56-59, 117, 150-157, 167, 208, 215, 220-227, 278
src/mysqlprovider.py      29     22    24%   18-21, 30-51
src/mysqlserver.py        57      4    93%   52-55
----------------------------------------------------
TOTAL                    212     42    80%
```

.gitignore
```diff
@@ -14,7 +14,7 @@ dist/
 downloads/
 eggs/
 .eggs/
-lib/
+# lib/
```
dstathis, line 17: You should remove this rather than comment it out

run_tests
```diff
@@ -14,4 +14,5 @@ fi
 
 flake8 || exit $?
 black -l 79 --diff --check src tests || exit $?
-python3 -m unittest -v "$@"
+# python3 -m unittest -v "$@"
```
dstathis, line 17: Again this line should just be removed

dstathis (commented):
Just a couple extraneous comments. Other than that it looks good.

.gitignore
Abuelodelanada: Done

run_tests
Abuelodelanada: done

src/mysqlserver.py
```diff
@@ -0,0 +1,92 @@
...
+            raise e
+
+    def is_ready(self) -> bool:
+        """Returns if MySQL is up and running"""
+        ready = False
+
+        try:
+            client = self._get_client()
+            ready = True
+            if client is not None:
```
balbirthomas, line 42: Feels a bit strange to be checking if client is not None after setting ready is True .

README.md
```diff
@@ -1,5 +1,8 @@
 # MySQL Operator
 
+![tests](https://github.com/canonical/mysql-operator/actions/workflows/run-tests.yaml/badge.svg)
```
balbirthomas, line 3: Nice. Note to myself to copy and use this. :-)

src/charm.py
```diff
@@ -53,6 +65,98 @@ def _on_peer_relation_changed(self, event):
...
+        }
+        return provides
+
+    # @property
+    def db_info(self) -> dict:
+        info = {
+            "app_name": self.model.app.name,
+            "host": self.hostname,
+            "port": self.model.config["port"],
+            "user_name": "root",
```
balbirthomas, line 155: I know I do this myself. I think we both should add some logic to return relation (client) specific username and passwords. There may be differences here for relational and non relational databases. For example if you have decided to let the client issue the query to create a database then the user created for the client will need to be have create privilege. On the other hand if you decide not to let the client create the data base then you charm lib `new_database` API may need an additional parameter, which is the database schema.

src/mysqlprovider.py
```diff
@@ -0,0 +1,53 @@
...
+        missing = None
+        if dbs_requested:
+            if dbs_available:
+                missing = list(set(dbs_requested) - set(dbs_available))
+            else:
+                missing = dbs_requested
+
+        if missing:
+            dbs_available.extend(missing)
+            logger.debug("SERVER REQUEST RESPONSE %s", dbs_available)
```
balbirthomas, line 50: Don't you need to do more here, to actually create the database ? Or is it the expectation that the client will issue the query to create the database ?

tests/test_charm.py
```diff
@@ -105,3 +109,107 @@ def test_new_unit_has_password(self):
...
+                isinstance(self.harness.charm.provides["provides"], dict)
+            )
+            self.assertEqual(
+                self.harness.charm.provides["provides"]["mysql"], VERSION
+            )
+            self.assertTrue(
+                isinstance(self.harness.charm.provides["config"], dict)
+            )
+            self.assertEqual(
+                self.harness.charm.provides["config"]["app_name"], "mysql"
```
balbirthomas, line 141: If you prefer, you can reduce multiple checks to a single one using `assertDictEquals`.

tests/test_charm.py
```diff
@@ -105,3 +109,107 @@ def test_new_unit_has_password(self):
...
+                self.harness.charm.provides["config"]["user_name"], "root"
+            )
+            self.assertEqual(
+                self.harness.charm.provides["config"]["mysql_root_password"],
+                "D10S!",
+            )
+
+    def test__on_start(self):
+        self.harness.set_leader(False)
+        self.assertEqual(self.harness.charm.on.start.emit(), None)
```
balbirthomas, line 160: I don't understand why you need to check the return value of `start.emit()`.

tests/test_charm.py
```diff
@@ -105,3 +109,107 @@ def test_new_unit_has_password(self):
             config["MYSQL_ROOT_PASSWORD"],
             self.harness.charm._stored.mysql_setup["MYSQL_ROOT_PASSWORD"],
         )
+
+    def test_provides(self):
```
balbirthomas, line 113: Just a stylistic suggestion, so you may choose to ignore if you do not like the suggestion. It will be great if tests names are a bit more descriptive. For example "test_version_and_db_info_is_provided" or anything else you think appropriate.

tests/test_charm.py
```diff
@@ -105,3 +109,107 @@ def test_new_unit_has_password(self):
...
+            config = {
+                "MYSQL_ROOT_PASSWORD": "D10S!",
+            }
+            self.harness.update_config(config)
+            self.harness.charm.on.start.emit()
+            self.assertEqual(
+                type(self.harness.charm.unit.status), ActiveStatus
+            )
+
+    def test__update_status(self):
```
balbirthomas, line 183: Another stylistic suggestion. Seems to me there are really three different test cases in this one test.
- Test non leader unit always sets active status with no message
- Test leader unit sets some message on failed status
- Test leader unit sets some other message on success status.
By splitting the tests into three separate tests has IMHO the following benefits
- Tests are smaller
- Tests names can be chosen to be very descriptive
- Failure of a test immediately indicates what is failing because the tests are specific and the names are descriptive.

balbirthomas (approved):
Approving so as to unblock you. But some of the comments requires a bit of thinking on both our parts.

src/charm.py
```diff
@@ -53,6 +65,98 @@ def _on_peer_relation_changed(self, event):
...
+            return
+
+        self._on_update_status(event)
+        self._stored.mysql_initialized = True
+        self.unit.status = ActiveStatus()
+
+    # Handles update-status event
+    def _on_update_status(self, event):
+        """Set status for all units
+        # FIXME
```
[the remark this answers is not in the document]
Abuelodelanada: Fixed

src/mysqlserver.py
Abuelodelanada: Fixed

src/mysqlprovider.py
Abuelodelanada: Sure, this part is in a WIP status

tests/test_charm.py
Abuelodelanada: Nice!
Done

tests/test_charm.py
Abuelodelanada: Yes, it is not clear, I will add comments in the test to clarify this

tests/test_charm.py
Abuelodelanada: Agree

tests/test_charm.py
Abuelodelanada: I agree with you.
done

src/charm.py
balbirthomas: I have pushed a prototype of client specific credentials to my [mongodb-operator branch](https://github.com/balbirthomas/mongodb-operator/tree/mongodb-charm-libs).

Abuelodelanada: Looks good @balbirthomas I'm going to start implementing something like this

PR #36, opened 2021-04-22
MySQL Lib

Abuelodelanada:
- [ ] Is provider-consumer-lib merged into ops? [PR502](https://github.com/canonical/operator/pull/502). Now we are using: `git+https://github.com/balbirthomas/operator/@provider-consumer-lib#egg=ops` Issue #37 
-  [ ] Is resource-oci-image ready? Now we are using: `git+https://github.com/juju-solutions/resource-oci-image/@c5778285d332edf3d9a538f9d0c06154b7ec1b0b#egg=oci-image` Issue #40 
- [ ] Change the use of hostname for ip. Issue #38 
- [ ]  Get `LIBID` from charmhub. Issue #39


lib/charms/mysql/v1/mysql.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def new_database(self):
+        """
+        Request creation of an additional database
+        """
+        if not self.charm.unit.is_leader():
+            return
+
+        rel_id = super()._stored.relation_id
+        if rel_id:
```
balbirthomas, line 55: @Abuelodelanada Apologies I should have emphasised, after discussion with @jlounder, it was decided to make rel_id an optional argument to this and related methods. If rel_id is provided use it, if not call get_relation without it.

src/charm.py
```diff
@@ -53,21 +58,107 @@ def _on_peer_relation_changed(self, event):
...
+                "MYSQL_ROOT_PASSWORD"
+            ],
+        }
+        return MySQL(mysql_config)
+
+    @property
+    def hostname(self) -> str:
+        """Unit hostname"""
+        unit_id = self.unit.name.split("/")[1]
+        return "{0}-{1}.{0}-endpoints".format(self.model.app.name, unit_id)
```
balbirthomas, line 138: Are you sure about this ? I may have forgotten but I think the third format string should be model name (kubernetes namespace) and not application name. Perhaps you set the two to be the same. Worth checking!.

src/charm.py
```diff
@@ -53,21 +58,107 @@ def _on_peer_relation_changed(self, event):
...
         if not self.unit.is_leader():
-            return
+            return None
 
         if "MYSQL_ROOT_PASSWORD" not in self._stored.mysql_setup:
-            password = "".join(
-                random.choice(ascii_letters + digits) for x in range(20)
-            )
-            self._stored.mysql_setup["MYSQL_ROOT_PASSWORD"] = password
+            self._stored.mysql_setup[
```
balbirthomas, line 159: Minor, stylistic issue : I do not see why all caps is necessary for a dictionary key. If this was a global constant I could see why but this does not seem to be a global or constant.

tests/check_coverage.py
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env python3
+# Copyright 2021 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""
+Very simple module to check unit tests coverage.
```
balbirthomas, line 6: Nice !

balbirthomas (approved):

src/charm.py
Abuelodelanada: I'll check it!
I created an issue related to this method #38

src/charm.py
Abuelodelanada: I used all caps just to follow the style in mysql docker image config options: https://hub.docker.com/r/ubuntu/mysql
Anyway I can change it

src/mysqlprovider.py
```diff
@@ -0,0 +1,121 @@
+#!/usr/bin/env python3
+# Copyright 2021 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""MySQLProvider module"""
```
dstathis, line 5: I think it would be best to put the provider class in the charmlib. That way, if any other charm wants to implement this interface, they can easily.

dstathis (approved):

src/mysqlprovider.py
balbirthomas: My understanding is that the charmlib provided by a charm is a facility for Consumers of the charm.

src/charm.py
balbirthomas: Your call. I appreciate that environment variables are typically all caps, but no reason you need to use the same capitalization in the environment variable and the dictionary key. Again completely your decision. Just my two cents :-).

src/charm.py
Abuelodelanada: In mongodb charm we are using:

```python
        return "{}-{}.{}-endpoints".format(self.app_name,
                                           id,
                                           self.app_name)
```

balbirthomas: Eeew. I need to check if I made the same mistake. Will do. Thanks. But if we are migrating to pebble and have to use IPs then I may not bother.

lib/charms/mysql/v1/mysql.py
Abuelodelanada: Done

src/charm.py
Abuelodelanada: We'll follow this in #41

PR #58, opened 2021-06-25
Actions for the MySQL charm

Abuelodelanada:


src/charm.py
```diff
@@ -75,6 +87,60 @@ def _on_update_status(self, event):
...
+        creds = {
+            "username": event.params["username"],
+            "password": event.params["password"],
+            "hostname": "%",
+        }
+
+        try:
+            self.mysql.set_user_password(creds)
+            event.set_results({"username": creds["username"]})
+            event.log(f"Pasword fo username: {creds['username']} changed")
```
balbirthomas, line 117: Password **for**

balbirthomas (approved):

src/charm.py
```diff
@@ -75,6 +87,60 @@ def _on_update_status(self, event):
...
+    def _on_create_user_action(self, event):
+        """Handle the create_user action."""
+        creds = {
+            "username": event.params["username"],
+            "password": event.params["password"],
+            "hostname": "%",
+        }
+
+        try:
+            self.mysql.new_super_user(creds)
```
mmanciop, line 99: Will this actually delete the previous user?

mmanciop (commented):
It is unclear to me what happens to an old superuser after creating a new one

src/charm.py
Abuelodelanada: No, this action does not delete previous users.
This action verifies if the username is already taken. If so will show `"User already exists: {credentials['username']}"`

Abuelodelanada:
> It is unclear to me what happens to an old superuser after creating a new one

At the end this action creates the new user and runs: `GRANT ALL PRIVILEGES ON *.* TO 'username'@'%'`

PR #68, opened 2022-01-04
Adds initial event handlers, placeholders, and metadata for builds

DnPlas:
This commit introduces the following:
* metadata for building the charm
* initial charmcraft.yaml configuration
* on_pebble_ready event handlers for mysql server and router
* placeholders for other event handlers
* initial tests


config.yaml
```diff
@@ -1,14 +1,15 @@
...
     type: string
+    default: ''
+    descriptions: Password for root user, equivalent to MYSQL_ROOT_PASSWORD.
+  mysql_user:
+    type: string
+    default: ''
+    description: MySQL username for privileged superuser. This is used in conjunction with MYSQL_PASSWORD. Equivalent to MYSQL_USER.
+  mysql_password:
+    type: string
+    description: Password for MYSQL_USER user.
```
simskij, line 15: This works as an intermediary way of configuring it, but once you've gotten a bit further in the implementation, it should probably be generated, and retrieved by the operator using an action.

metadata.yaml
```diff
@@ -1,25 +1,26 @@
...
+  Kubernetes charm for HA MySQL InnoDB Clusters.
 
-# TEMPLATE-TODO: replace with containers for your workload (delete for non-k8s)
 containers:
-  httpbin:
-    resource: httpbin-image
+  mysql-server:
+    resource: mysql-server-image
+  mysql-shell:
+    resource: mysql-shell-image
```
simskij, line 14: Isn't this already included in the server image?

simskij (commented):
🎉 🥳

config.yaml
quiet-ranger: Passwords stored in plain text in a file are considered bad practice. Generating them at deployment time will be safer (the option to query must be provided).  One important consideration is that with every deployment a new password would be generated and this could be a problem for some enterprises.  One step further would be to generate upon first deployment, store in a secure repository (Juju secrets?) and reuse on subsequent deployments.  Yet another consideration should be when a company has a strict password expiry policy.

mmanciop: The pattern we use in the O11y team is to have the operator automatically generate the passwords, and provide and action for the administrator to retrieve them

DnPlas: The intention is to default these passwords to an empty string and have the user change that through config, when a ConfigChanged is triggered, these defaults will be wiped away.

metadata.yaml
DnPlas: No, the server image has only the server binary.

simskij: https://hub.docker.com/r/mysql/mysql-server/
https://hub.docker.com/r/ubuntu/mysql

both include the shell. are you using some other image?

config.yaml
simskij: > The intention is to default these passwords to an empty string and have the user change that through config

Thanks for the clarification! 🙏🏽 

What I'm suggesting is that these defaults probably should be generated as well, and that setting them through config values likely isn't something we want in the long-run.

See https://github.com/canonical/grafana-k8s-operator/blob/e3b5795cb3ff0da328549b3e266bfb8c1ebb4047/src/charm.py#L539-L562 for reference. If/Once the user sets them to something else (in this case using for instance the mysql-cli), the action to fetch the default password would be disabled.

DnPlas: I absolutely agree with generating random defaults, that's better. Will change that.

marceloneppel (commented):
I only saw the point about the password that Juan also commented on. Other than that, LGTM.

config.yaml
DnPlas: Thanks for the suggestions. I've removed the `config.yaml` and added a new helper method in the charm code to generate a random password.

paulomach:
@DnPlas LGTM, but should we fix then first?

src/charm.py
```diff
@@ -1,104 +1,126 @@
...
+        current_layer = container.get_plan()
+        # Create a new config layer
+        service = container.name
+        new_layer = self._mysql_layer(service=service)
+
+        if current_layer.services != new_layer.services:
+            container.add_layer(service, new_layer, combine=True)
+            logging.info("Pebble plan updated with new configuration")
+            if not container.get_service(service).is_running():
+                container.restart(service)
```
marceloneppel, line 62: Maybe this could be replaced by a `container.start`?

DnPlas: Probably the condition is not even needed and I can just leave the `container.restart()`.

marceloneppel: I agree.

DnPlas: Just changed it, thanks.

src/charm.py
```diff
@@ -1,104 +1,123 @@
...
+    def _on_pebble_ready(self, event: WorkloadEvent) -> None:
+        """Updates the Pebble configuration layer when changed."""
         container = event.workload
-        # Define an initial Pebble layer configuration
-        pebble_layer = {
-            "summary": "httpbin layer",
-            "description": "pebble config layer for httpbin",
+        if not container.can_connect():
+            self.unit.status = WaitingStatus("Waiting for pod startup to complete")
+            return
```
quiet-ranger, line 50: Am I wrong if I say that this return should be changed to a defer?  You only receive the pebble ready event once so, if you don't execute the lines that follow this point they would never be executed.

marceloneppel (commented):
Only that change about restart, which has already been fixed. LGTM!

src/charm.py
DnPlas: Updated, thanks @quiet-ranger

tests/unit/test_charm.py
```diff
@@ -4,64 +4,39 @@
...
+
+                # Test WaitingStatus if we can't connect to the container
+                with patch("ops.model.Container.can_connect") as _can_connect:
+                    _can_connect.return_value = False
+                    self.harness.container_pebble_ready(service)
+                    self.assertEqual(
+                        self.harness.model.unit.status,
+                        WaitingStatus("Waiting for pod startup to complete"),
+                    )
+
```
MiaAltieri, line 40:
```suggestion
# Test pebble_ready when we can connect to the container
```

tests/unit/test_charm.py
```diff
@@ -4,64 +4,39 @@
...
-        # Get the plan now we've run PebbleReady
-        updated_plan = self.harness.get_container_pebble_plan("httpbin").to_dict()
-        # Check we've got the plan we expected
-        self.assertEqual(expected_plan, updated_plan)
-        # Check the service was started
-        service = self.harness.model.unit.get_container("httpbin").get_service("httpbin")
-        self.assertTrue(service.is_running())
-        # Ensure we set an ActiveStatus with no message
-        self.assertEqual(self.harness.model.unit.status, ActiveStatus())
+    def test_on_pebble_ready(self):
```
MiaAltieri, line 21: You'll want to [this scenario](https://github.com/canonical/mysql-k8s-operator/blob/4212fd595b7738a8a7f2dcada5d7e8a59e0eaa65/src/charm.py#L59-L62) from `charm.py` specifically when `current_layer.services != new_layer.services` and `current_layer.services == new_layer.services`

MiaAltieri (commented):
This looks great! Some minor nits and questions. Also you may want to increase your unit test coverage or save it for a future PR

metadata.yaml
```diff
@@ -1,25 +1,31 @@
 # Copyright 2021 Canonical Ltd.
 # See LICENSE file for licensing details.
 
-# For a complete list of supported options, see:
-# https://discourse.charmhub.io/t/charm-metadata-v2/3674/15
-name: operator-template
-display-name: |
```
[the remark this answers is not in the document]
DnPlas: Since we are using Charmhub now, is this necessary?

src/charm.py
```diff
@@ -1,104 +1,124 @@
...
-            "description": "pebble config layer for httpbin",
+        if not container.can_connect():
+            self.unit.status = WaitingStatus("Waiting for pod startup to complete")
+            event.defer()
+            return
+
+        # Get current config
+        current_layer = container.get_plan()
+        # Create a new config layer
+        service = container.name
```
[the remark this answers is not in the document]
DnPlas: Well, `services` is an attribute of `Layer` whilst `service` is the name of a variable, but I guess I can improve readability changing `service` to `service_name`.

src/charm.py
```diff
@@ -1,104 +1,124 @@
...
+
+        if current_layer.services != new_layer.services:
+            container.add_layer(service, new_layer, combine=True)
+            logging.info("Pebble plan updated with new configuration")
+            container.restart(service)
+
+        self.unit.status = ActiveStatus()
+
+    # ---- Helper functions
+    def _generate_password(self) -> str:
```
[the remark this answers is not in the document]
DnPlas: We can have one-line docstrings if the method return is obvious.

src/charm.py
```diff
@@ -1,104 +1,124 @@
...
+        """Returns a random 12 alphanumeric string that can be used as password."""
+        alphanums = string.ascii_letters + string.digits
+        return "".join(secrets.choice(alphanums) for _ in range(12))
+
+    def _mysql_layer(self, service: str) -> Layer:
+        """Returns a pre-configured Pebble Layer."""
```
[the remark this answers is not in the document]
DnPlas: Yeh, it applies here.

src/charm.py
```diff
@@ -1,104 +1,124 @@
...
+                        "startup": "enabled",
+                        "environment": env_config,
+                    }
+                },
+            }
+        return Layer(layer_config)
+
+    # ---- Properties
+    @property
+    def config(self):
```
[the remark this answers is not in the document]
DnPlas: I agree on private methods for this.

tests/unit/test_charm.py
DnPlas: I'm sorry, I don't get your comment, could you please elaborate?

tests/unit/test_charm.py
```diff
@@ -4,64 +4,39 @@
...
+                with patch("ops.model.Container.can_connect") as _can_connect:
+                    _can_connect.return_value = False
+                    self.harness.container_pebble_ready(service)
+                    self.assertEqual(
+                        self.harness.model.unit.status,
+                        WaitingStatus("Waiting for pod startup to complete"),
+                    )
+
+                self.harness.container_pebble_ready(service)
+                self.assertEqual(self.harness.model.unit.status, ActiveStatus())
```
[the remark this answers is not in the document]
DnPlas: Yeah, since this is a super basic PR, I just added the most basic tests to make the CI pass, but you are right, I've to add more in the future.

metadata.yaml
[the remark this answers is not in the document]
MiaAltieri: Ah you're right! I am not sure 🤷

tests/unit/test_charm.py
MiaAltieri: Ah I had a typo in it so it didn't make sense, I corrected it but let me try to be more clear 😅 

I was trying to say that the current test `test_on_pebble_ready ` tests [this part of the code](https://github.com/canonical/mysql-k8s-operator/blob/4212fd595b7738a8a7f2dcada5d7e8a59e0eaa65/src/charm.py#L45-L57) and the Active Status part. But I don't think it tests the case where the `layer.services` differ, ie [this part](https://github.com/canonical/mysql-k8s-operator/blob/4212fd595b7738a8a7f2dcada5d7e8a59e0eaa65/src/charm.py#L59-L63) and that you might want to test it to be sure it adds a layer when necessary.

DnPlas: Ah, I get it. Although we have to, I'm not testing the expected `Layer` vs the actual layer (like in Postgres, even in Mongo I think) at least in this PR. I left it like that because I'm pretty sure I'll change the layer definitions, and I didn't want to add a test that I'll have to change later. I'm planning to do something similar to what @marceloneppel did for [Postgres](https://github.com/canonical/postgresql-k8s-operator/blob/main/tests/unit/test_charm.py#L78).

metadata.yaml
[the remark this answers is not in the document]
DnPlas: Okay, seems like it is currently [not supported](https://github.com/juju/charm/blob/master/meta.go#L259-L282).

src/charm.py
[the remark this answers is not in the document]
DnPlas: Updated the docstring.

src/charm.py
```diff
@@ -1,104 +1,124 @@
...
+        current_layer = container.get_plan()
+        # Create a new config layer
+        service = container.name
+        new_layer = self._mysql_layer(service=service)
+
+        if current_layer.services != new_layer.services:
+            container.add_layer(service, new_layer, combine=True)
+            logging.info("Pebble plan updated with new configuration")
+            container.restart(service)
+
```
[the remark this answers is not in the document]
DnPlas: Yeh, I agree that something could go wrong with the service restart. It could be something like:

```python
try:
    container.restart(service_name)
except ops.pebble.ChangeError as e:
    <something, maybe BlockedStatus?>
else:
```

I haven't seen examples of error handling for the restart, so I'm not sure. @jnsgruk thoughts?

metadata.yaml
```diff
@@ -1,25 +1,31 @@
...
-  TEMPLATE-TODO: fill out a display name for the Charmcraft store
+name: mysql-k8s
 description: |
-  TEMPLATE-TODO: fill out the charm's description
+  MySQL RDBMS with InnoDB Cluster support for Kubernetes.
 summary: |
-  TEMPLATE-TODO: fill out the charm's summary
+  Kubernetes charm for HA MySQL InnoDB Clusters.
```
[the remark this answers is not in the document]
DnPlas: Well, summaries are usually one liners.

paulomach:
@shayancanonical @delgod @quiet-ranger 

What if we merge this, since it establish the basis of a working (even though not functional) charm, to then discuss in depth and review the specs ?

src/charm.py
```diff
@@ -1,104 +1,129 @@
...
+        self.framework.observe(self.on.start, self._on_start)
         self.framework.observe(self.on.config_changed, self._on_config_changed)
-        self.framework.observe(self.on.fortune_action, self._on_fortune_action)
-        self._stored.set_default(things=[])
+        change_events = [
+            self.on.mysql_server_pebble_ready,
+            self.on.mysql_router_pebble_ready,
+        ]
+        for event in change_events:
+            self.framework.observe(event, self._on_pebble_ready)
```
pengale, line 32: Style nitpick: I think that you can flatten this loop with:

```
# Pebble Ready Handlers
self.framework.observe(self.on.mysql_server_pebble_ready, self._on_pebble_ready)
self.framework.observe(self.on.mysql_router_pebble_ready, self._on_pebble_ready)
```

That makes it visually clear that these are part of the block of calls to self.framework.observe, and is in keeping with our example code. The comment helps the reader understand that there might be something to pay attention to (in this case, attaching a handler to two events).

pengale: ... if you wanted to, you could setup a struct mapping events to handlers, and call self.framework.observe in a loop through the map. But that's not what we've done elsewhere, and I think that being simple wins out over being DRY here.

src/charm.py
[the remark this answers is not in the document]
pengale: The answer to that question depends on whether the failure is something that you can recover automatically from.

If you can recover, execute the recovery code (which might just look like deferring the event, possibly tracking how many times that you have done so in self._stored, so that you can give up after some time).

If you can't recover, and a human needs to intervene, set BlockedStatus and return.

Note that if you don't catch the error, the hook will fail, and the charm will go into error status, re-running the hook until the problem fixes itself, or a human intervenes.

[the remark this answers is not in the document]
pengale: Note: implicit in the answer above is that you have some understanding of the underlying workload. What can go wrong when you're setting up MySQL? What sort of errors will you get back? Is the service ready immediately after the restart, or do you need to do a liveness check? 

This is the heart of what charms are: a framework that makes it easy to focus just on the bits unique to the workload, with code that handles common failure modes, and best practices around handling those failures.

src/charm.py
jnsgruk: +1 on flattening this out - not a big issue, but I think it'll be nicer to read.

paulomach: Can't disagree with that. @DnPlas , by now you can let any further change for @shayancanonical or myself

simskij:
@shayancanonical @DnPlas, should we close this PR then?

shayancanonical:
Since there are significant differences between @paulomach's implementation and this PR, I am going to close this PR. Thanks @DnPlas - we really benefited from using this PR as a reference

PR #69, opened 2022-02-14
Adds code to configure MySQL instances and create cluster

DnPlas:
WIP


src/charm.py
```diff
@@ -19,30 +21,48 @@
...
     # ---- Event handlers
-    def _on_start(self, _):
-        """Placeholder for StartEvent."""
-        pass
+    def _on_install(self, event: WorkloadEvent) -> None:
+        """Ensures the MySQL server instance is configured for an InnoDB Cluster."""
+        container = self.unit.get_container(self._server_container)
+        if not container.can_connect():
+            event.defer()
+            return
```
pengale, line 49: It looks like you're doing most of your install tasks during your _pebble_ready handler. I think that you can simply leave the _on_install handler out of this charm.

src/charm.py
```diff
@@ -19,30 +21,48 @@
...
+        """Set all needed passwords for connection and administration."""
+        self._stored.root_password = self._root_password
+        self._stored.mysql_password = self._mysql_password
 
     def _on_pebble_ready(self, event: WorkloadEvent) -> None:
         """Updates the Pebble configuration layer when changed."""
         container = event.workload
+
+        # QUESTION: if you check this and defer on_install,
+        # is it worth it to check it here as well?
```
pengale, line 65: Yes. You should always check container.can_connect in code that assumes that you can connect to the container. There is no contract that guarantees that .can_connect will continue to return True, even if you have seen it return True in the past.

src/charm.py
```diff
@@ -19,30 +21,48 @@
...
-        """Placeholder for ConfigChangedEvent."""
-        pass
+        # Configure server instance:
+        # Push configuration file and instance script to workload container
+        self._mysql.configure_server_instance()
+
+    def _on_leader_elected(self, event: WorkloadEvent) -> None:
+        """Set all needed passwords for connection and administration."""
+        self._stored.root_password = self._root_password
+        self._stored.mysql_password = self._mysql_password
```
pengale, line 58: Since this information lives in the config, you should not track it in the StoredState. 

The Operator Framework loads an up-to-date copy of the config when it starts, at the beginning of each invocation. You can rely on that data being accurate, and low cost to access.

src/charm.py
```diff
@@ -116,14 +138,33 @@ def _mysql_layer(self, service_name: str) -> Layer:
...
     def _config(self):
-        """Configuration data for MySQL authentication."""
+        """Configuration data for MySQL."""
         config = {
+            "innodb_cluster_name": self.model.config["innodb_cluster_name"],
             "mysql_user": "mysql_user",
-            "mysql_password": self._generate_password(),
-            "mysql_root_password": self._generate_password(),
+            "mysql_password": self._mysql_password,
+            "mysql_root_password": self._root_password,
```
pengale, line 146: As above. Get this from self.model.config['mysql_password'] and self.model.config['root_password']. This is a much more reliable source of truth than the StoredState.

src/charm.py
```diff
@@ -116,14 +138,33 @@ def _mysql_layer(self, service_name: str) -> Layer:
...
         }
         return config
 
+    @property
+    def _root_password(self):
+        root_password = self._stored.root_password
```
pengale, line 152: I like this encapsulation! But you should get the root_password from the config, rather than the StoredState.

config.yaml
```diff
@@ -1,14 +1,7 @@
...
-# If you don't need a config, you can remove the file entirely.
-# It ties in to the example _on_config_changed handler in src/charm.py
-#
-# Learn more about config at: https://juju.is/docs/sdk/config
-
 options:
-  thing:
-    default: 🎁
-    description: A thing used by the charm.
+  innodb_cluster_name:
```
pengale, line 4: Since this seems to be something that can be set at runtime, it makes sense to put this in the config, and access it with `self.model.config['innodb_cluster_name']`. That gives operators a nice transparent way to set it.

pengale (commented):
Thank you for sharing this Draft -- I really appreciate being able to see and comment on Work In Progress code.

I left some comments! Please feel free to ask follow questions, either here or on mattermost. (I'm @pengale) in both places.

src/charm.py
jnsgruk: Hmm, I think probably not in this case. Having passwords in config makes them a little tooooo easy to reveal.

A common pattern is autogenerate a root pw, and have an action to retrieve that initial password (generally stored on the peer relation) 

(We need Juju secrets...!)

pengale: The password is already stored in config in this draft. I agree that, without a released implementation of Juju Secrets, this is not a great practice.

Having the leader generate and share the password is a more secure interim practice. We'd only want to cache the password's value in StoredState if the value is unique to reach unit, and we don't mind it changing whenever the pod is rescheduled.

simskij:
@DnPlas @shayancanonical, should we close this PR then?

shayancanonical:
Since there are significant differences between @paulomach's implementation and this PR, I am going to close this PR. Thanks @DnPlas - we really benefited from using this PR as a reference

PR #72, opened 2022-04-12
Replicate mysql-helpers from machine charm

paulomach:
# Issue
Replication of machine charm code into k8s. 
Ticket is [DPE-204](https://warthogs.atlassian.net/browse/DPE-204)

# Solution
Necessary to make adaptations in order to use `container.exec` calls instead of `subprocess` ones.
**NOTE**: Event though we made decision about how to manage shared code, it wont be done unitl we have HA working on both machine and k8s charm.

# Context
Machine charm is driving the development and code must be replicated in k8s charm.

# Testing
Test is initially implemented in unit tests, also adapted from the machine charm.
**NOTE**: integration testing is not implemented due to the charm not implementing any functionality yet. This will be the follow up PR.

# Release Notes
* add `mysqlsh_helpers`
* add `test_mysqlsh_helpers` with unit tests
* bump ops version + add tenacity for retrying things


src/mysqlsh_helpers.py
```diff
@@ -0,0 +1,176 @@
...
+    def configure_mysql_users(self) -> None:
+        """Configure the MySQL users for the instance.
+
+        Creates base `root@%` and `clusteradmin@instance_address` user with the
+        appropriate privileges, and reconfigure `root@localhost` user password.
+        Raises MySQLConfigureMySQLUsersError if the user creation fails.
+        """
+        _script = (
+            "SET @@SESSION.SQL_LOG_BIN=0;",
+            f"CREATE USER '{self.cluster_admin_user}'@'{self.instance_address}' IDENTIFIED BY '{self.cluster_admin_password}';",
```
shayancanonical, line 83: two comments:

1. we should create the `serverconfig@%` user manually, and then use `serverconfig@%` user to call `dba.configure_instance`. it is while configuring the instance that we create the `clusteradmin@%` user by passing in `clusterAdmin` and `clusterAdminPassword` parameters to `configure_instance`
2. we should avoid using `mysql_native_password` -> to do this, we may need to break the commands into 2 groups of commands (still to be verified by Mykola in the mysql-operator repository)

src/mysqlsh_helpers.py
```diff
@@ -0,0 +1,176 @@
...
+        except ExecError as e:
+            logger.exception(f"Failed to configure users for: {self.instance_address}", exc_info=e)
+            raise MySQLConfigureMySQLUsersError(e.stdout)
+
+    def configure_instance(self) -> None:
+        """Configure the instance to be used in an InnoDB cluster."""
+        commands = (
+            f"dba.configure_instance('{self.cluster_admin_user}:{self.cluster_admin_password}@{self.instance_address}')",
+            f"my_shell = shell.connect('{self.cluster_admin_user}:{self.cluster_admin_password}@{self.instance_address}')",
+            'my_shell.run_sql("RESTART;");',
```
shayancanonical, line 105: i ran into an issue that `RESTART` tries to restart with systemd. we'd need to confirm this is the case, and if so, we'd need to restart with pebble

src/mysqlsh_helpers.py
paulomach: I guess I was waiting for the merge in machine land, but I can use the branch code anyway right.

src/mysqlsh_helpers.py
paulomach: That's a miss, well remembered.

src/mysqlsh_helpers.py
```diff
@@ -0,0 +1,176 @@
...
+
+        Args:
+            container: Container to run the script in
+
+        Returns:
+            Path to binary mysqlsh
+        """
+        # Allow for various versions of the mysql-shell snap
+        # When we get the alias use /snap/bin/mysqlsh
+        _paths = ("/snap/bin/mysqlsh", "/snap/bin/mysql-shell.mysqlsh")
```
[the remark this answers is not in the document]
paulomach: Yep, you're right. Done on 1abc440

paulomach:
I decided to close this one and re-submit it after a rebase from the [PR #6](https://github.com/canonical/mysql-operator/pull/6) of the machine charm, instead of the main, since it already fix many of the comments in this PR.

PR #73, opened 2022-04-14
Replicate mysql helpers

paulomach:
# Issue
Replication of machine charm helpers (MySQL abstraction classes) for k8s. See [this merged PR](https://github.com/canonical/mysql-operator/pull/6) for full context.
Ticket is [DPE-204](https://warthogs.atlassian.net/browse/DPE-204), but also [DPE-140](https://warthogs.atlassian.net/browse/DPE-140)

# Solution
Necessary to make adaptations in order to use `container.exec` calls instead of `subprocess` ones and differences on mysql setup due to some configuration done in the `entrypoint.sh` script provided by the [base oci image](https://hub.docker.com/r/ubuntu/mysql).
Differences are:
* calls to mysql and mysqlsh at `_run_mysqlcli_script` and `_run_mysqlsh_script` functions
* `configure_users_commands` sql script at `configure_mysql_users` function
* different `options` at `configure_instance` function

**NOTE**: Event though we made decision about how to manage shared code, it wont be done unitl we have HA working on both machine and k8s charm.

# Context
Machine charm is driving the development and code must be replicated in k8s charm.

# Testing
Tests implemented in unit tests, refactored from the machine charm.
Additional testing done locally with prototype `charm.py` code (see note bellow).
**NOTE**: integration testing is not implemented due to the charm not implementing any functionality yet. This will be the follow up PR.

# Release Notes
* add `mysqlsh_helpers`
* add `test_mysqlsh_helpers` with unit tests
* bump ops version + add tenacity for retrying things

PR #74, opened 2022-04-26
From zero to HA

paulomach:
# Issue

Mysql-k8s charm was not doing much till this PR.
This PR brings mysql-k8s up to date with mysql-machine charm i.e. deployable charm that can scale up in a cluster setting.

# Solution

In short, implemented charm lifecycle as in [this in repo flow](https://github.com/canonical/mysql-k8s-operator/blob/feature/peer_relation_data/CONTRIBUTING.md#appendix).

## In long:
Implemented charm handlers for:
* leader_elected - set configuration
* pebble_ready - bootstrap/configure instance & users, and if leader, create the cluster
* config_changed - ensure config
* peer_relation_joined - add non-leader to cluster
 
Implemented charm action for getting generated passwords for users `"root-password", "server-config-password", "cluster-admin-password"`.
And all surrounding boiler plate, most unashamedly copied over the machine charm. 

# Testing

## Unit
Tests are implemented in unit testing for:
* [tests/unit/test_charm.py](https://github.com/canonical/mysql-k8s-operator/blob/feature/peer_relation_data/tests/unit/test_charm.py) - tests for the charm - basic status transitions and helpers
* [tests/unit/test_mysqlsh_helpers.py](https://github.com/canonical/mysql-k8s-operator/blob/feature/peer_relation_data/tests/unit/test_mysqlsh_helpers.py) - tests for helpers functions - extended from previous PR #73 

## Integration
Integration tests that execute, in order:
1. single node cluster deployment
2. scale up cluster to 3 nodes (e.g. HA)
3. test write on primary node and read content on secondary node

# Release Notes
* .github/workflows/ci.yaml - Add configuration to CI (need to build custom OCI from [canonical/mysql-container](https://github.com/canonical/mysql-container)
* CONTRIBUTING.md - Add fancy flow representation of implemented handlers
* actions.yaml - Add action to retrieve generated passwords
* config.yaml - Add optional `cluster-name` config
* metadata.yaml - Add storage, peer relationship and fixed image name
* pyproject.toml - Add D411 to ignore since was giving false positives
* src/charm.py - Implemented lifecycle from zero to scaling up cluster
* src/mysqlsh_helpers.py - Add new functions, namely:
  * bootstrap_instance - bootstrap mysqld manually (ditch the usage of container entrypoint script)
  * patch_dns_searches - add searches to workaround mysqld limitations with name resolution in k8s environment
  * is_instance_configured_for_innodb - test if a given instance is ready to be added to cluster
  * plus modifications on the users initial config/creation, in response to ditching the usage of container entrypoint
* tests/integration/test_charm.py - Integration tests as described in previous section
* tests/unit/__init__.py - Add recommended flag
* tests/unit/test_charm.py - Unit tests as described in previous section
* tests/unit/test_mysqlsh_helpers.py - Increased coverage and add new functions tests
* tox.ini - Added mysql-connector library used in tests


src/charm.py
```diff
@@ -4,27 +4,263 @@
...
+        # Only execute on unit leader
+        if not self.unit.is_leader():
+            return
+
+        # Set the cluster name in the peer relation databag if it is not already set
+        peer_data = self._peers.data[self.app]
+
+        if not peer_data.get("cluster-name"):
+            peer_data["cluster-name"] = (
+                self.config.get("cluster-name") or f"cluster-{generate_random_hash()}"
```
marceloneppel, line 135: The config name is different from the name on `config.yaml`.
```suggestion
                self.config.get("cluster_name") or f"cluster-{generate_random_hash()}"
```

src/charm.py
```diff
@@ -4,27 +4,263 @@
...
+
+        for required_password in required_passwords:
+            if not peer_data.get(required_password):
+                logger.debug(f"Setting {required_password}")
+                password = generate_random_password(PASSWORD_LENGTH)
+                peer_data[required_password] = password
+
+        if not peer_data.get("cluster-name"):
+            peer_data["cluster-name"] = (
+                self.config.get("cluster-name") or f"cluster-{generate_random_hash()}"
```
marceloneppel, line 155: Same as above.

The config name is different from the name on `config.yaml`. It will not get the config value this way.
```suggestion
                self.config.get("cluster_name") or f"cluster-{generate_random_hash()}"
```

tests/unit/test_charm.py
```diff
@@ -1,11 +1,166 @@
...
+
+        self.assertTrue(isinstance(self.charm.unit.status, BlockedStatus))
+
+    def test_on_config_changed(self):
+        # Test config changed set of cluster name
+        self.assertEqual(self.charm._peers.data[self.charm.app].get("cluster-name"), None)
+        self.harness.set_leader()
+        self.charm.on.config_changed.emit()
+        self.assertNotEqual(
+            self.charm._peers.data[self.charm.app].get("cluster-name"), "not_valid_cluster_name"
```
marceloneppel, line 111: I could not find this (maybe a) constant `not_valid_cluster_name` on the charm code.

tests/integration/test_charm.py
```diff
@@ -15,28 +17,115 @@
...
 
-    # effectively disable the update status from firing
-    await ops_test.model.set_config({"update-status-hook-interval": "60m"})
+
+@pytest.mark.skip_if_deployed
+@pytest.mark.abort_on_fail
+async def test_scale_to_ha(ops_test: OpsTest):
+    """Scale the charm to HA."""
+    await ops_test.model.applications[APP_NAME].scale(3)
+    await ops_test.model.wait_for_idle(apps=[APP_NAME], status="active", timeout=1000)
```
zmraul, line 53: You can avoid the `wait_for_idle` inconsistency issues by using something like:
```python
await ops_test.model.block_until(lambda: len(ops_test.model.applications[APP_NAME].units) == 3)
```

CONTRIBUTING.md
```diff
@@ -61,3 +61,41 @@ juju deploy ./mysql-k8s_ubuntu-20.04-amd64.charm \
 ## Canonical Contributor Agreement
 
 Canonical welcomes contributions to the MySQL Operator. Please check out our [contributor agreement](https://ubuntu.com/legal/contributors) if you're interested in contributing to the solution.
+
+## Appendix
+
+### Charm lifecycle flowcharts
```
zmraul, line 67: I think adding this is a good idea!

tests/integration/test_charm.py
```diff
@@ -15,28 +17,115 @@
...
+
+    assert len(ops_test.model.applications[APP_NAME].units) == 3
```
marceloneppel, line 55: Maybe it might be worth checking if they are part of the same cluster or something else to ensure they scale and provide HA in this test. I know the following test guarantees this, but I'm wondering if it's not better to test something else here to justify the test's name.

marceloneppel (commented):
LGTM! Nice work, Paulo! There are some minor details to check.

actions.yaml
```diff
@@ -1,2 +1,5 @@
-# Copyright 2021 Canonical Ltd.
+# Copyright 2022 Canonical Ltd.
 # See LICENSE file for licensing details.
+
+get-generated-passwords:
+  description: Get the cluster admin user password
```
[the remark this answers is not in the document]
paulomach: It should. done on 31449cb

config.yaml
```diff
@@ -0,0 +1,7 @@
+# Copyright 2022 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+options:
+  cluster_name:
+    description: "Optional - Name of the MySQL InnoDB cluster"
+    type: "string"
```
zmraul: Why generate a random cluster name on the charm instead of just having a default name here?

src/charm.py
```diff
@@ -4,27 +4,263 @@
...
+    random_characters = generate_random_password(10)
+    return hashlib.md5(random_characters.encode("utf-8")).hexdigest()
+
 
 class MySQLOperatorCharm(CharmBase):
     """Operator framework charm for MySQL."""
 
     def __init__(self, *args):
         super().__init__(*args)
+        self._mysqld_service = "mysqld"
```
zmraul, line 60: Since this is not changed, only accessed, it could be made a literal at the top. This is personal preference more than anything.

src/mysqlsh_helpers.py
```diff
@@ -264,9 +322,38 @@ def _run_mysqlsh_script(self, script: str) -> None:
             MYSQLSH_SCRIPT_FILE,
         ]
         process = self.container.exec(cmd)
-        process.wait_output()
+        stdout, _ = process.wait_output()
+        return stdout
```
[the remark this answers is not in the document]
paulomach: deb734a

src/charm.py
paulomach: fix on deb734a

src/charm.py
```diff
@@ -4,27 +4,263 @@
...
+                logger.debug("Unable to configure instance: {}".format(e))
+                return
+
+            if self.unit.is_leader():
+                try:
+                    # Create the cluster when
+                    self._mysql.create_cluster()
+                except MySQLCreateClusterError as e:
+                    self.unit.status = BlockedStatus("Unable to create cluster")
+                    logger.debug("Unable to create cluster: {}".format(e))
```
zmraul, line 220: Should this be followed by a return?

tests/unit/test_charm.py
paulomach: It's actually testing against the string `not_valid_cluster_name`. Generated cluster names will have the format `cluster-<10 chars hash>`, so the assert must pass when it differs. I've changed it on 8251bfe to ensure it fails when key is not present (by not using dict.get).

tests/integration/test_charm.py
paulomach: You're right. Will check that.

src/charm.py
```diff
@@ -4,27 +4,263 @@
...
 
+    def _on_peer_relation_joined(self, event: RelationJoinedEvent):
+        """Handle the peer relation joined event."""
+        # Only leader unit add instances to the cluster
+        if not self.unit.is_leader():
+            return
+
+        new_instance = self._get_hostname_by_unit(event.unit.name)
+
+        if not self._mysql.is_instance_configured_for_innodb(new_instance):
```
zmraul, line 250: I'm not sure if this is possible: the `peer_relation_joined` happens right after `peer_relation_created`, both during setup and before `leader_elected` event. Then calling `_mysql` would fail since it's not configured. A check for peer_data['configured'] would be needed.
As I said though, I don't know if the sequence of events can happen that way.

zmraul (commented):
Looks nice!
I didn't look all of it, but I left some comments.

config.yaml
paulomach: One can have multiple database clusters on the same k8s cluster - having no default can avoid collisions.

src/charm.py
paulomach: Same here, the reasoning was what others were doing, but I may reconsider.

src/charm.py
paulomach: Yes! 0fb998f

src/charm.py
paulomach: In any case, this will be mitigate by this:
```python
if not self.unit.is_leader():
            return
```
above the instance adding.

zmraul: Ah, I missed that :)

pyproject.toml
```diff
@@ -29,7 +30,7 @@ exclude = [".git", "__pycache__", ".tox", "build", "dist", "*.egg_info", "venv"]
 select = ["E", "W", "F", "C", "N", "R", "D", "H"]
 # Ignore W503, E501 because using black creates errors with this
 # Ignore D107 Missing docstring in __init__
-ignore = ["W503", "E501", "D107"]
+ignore = ["W503", "E501", "D107", "D411"]
```
shayancanonical, line 33: why did we add `D411` to the ignore list?

src/charm.py
```diff
@@ -4,27 +4,269 @@
...
+    """
+    choices = string.ascii_letters + string.digits
+    return "".join([secrets.choice(choices) for i in range(length)])
+
+
+def generate_random_hash() -> str:
+    """Generate a hash based on a random string.
+
+    Returns:
+        A length 10 hash based on a random string.
```
shayancanonical, line 49: i believe that this comment is misleading - the hash will not be length 10 (rather the seed to generate the random hash will have length 10)

pyproject.toml
paulomach: I'm getting a (apparent) false positive: did not give much investigation though

src/charm.py
```diff
@@ -4,27 +4,269 @@
...
+
+        # Add new instance to the cluster
+        self._mysql.add_instance_to_cluster(new_instance)
+        logger.debug(f"Added instance {new_instance} to cluster")
+
+    # =========================================================================
+    # Charm action handlers
+    # =========================================================================
+    def _get_generated_passwords(self, event: ActionEvent) -> None:
+        event.set_results(
```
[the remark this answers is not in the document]
paulomach: I have no strong position here either - this approach is to make my life easier now.

src/mysqlsh_helpers.py
```diff
@@ -86,6 +101,52 @@ def __init__(
...
+            app_name: name of the application
+        """
+        try:
+            resolv_file = self.container.pull("/etc/resolv.conf")
+            content = resolv_file.read()
+            output_string = ""
+            for line in content.splitlines():
+                line_content = line.split()
+                if line_content[0] == "search":
+                    line_content[1:1] = [f"{app_name}-endpoints.{line_content[1]}"]
```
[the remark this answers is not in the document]
paulomach: Nops!

tests/integration/test_charm.py
paulomach: I'm doing a big refactor on the test code. Wait for updates

src/charm.py
```diff
@@ -4,27 +4,269 @@
...
+
+        for required_password in required_passwords:
+            if not peer_data.get(required_password):
+                logger.debug(f"Setting {required_password}")
+                password = generate_random_password(PASSWORD_LENGTH)
+                peer_data[required_password] = password
+
+        if not peer_data.get("cluster-name"):
+            peer_data["cluster-name"] = (
+                self.config.get("cluster-name") or f"cluster-{generate_random_hash()}"
```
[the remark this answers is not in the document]
paulomach: `c-v,c-p` - fixed

src/charm.py
paulomach: fixed

pyproject.toml
paulomach: I've removed the offending thing completely.

src/charm.py
paulomach: Change as suggested

paulomach (commented):
Addressed PR comments

src/mysqlsh_helpers.py
```diff
@@ -15,18 +15,25 @@
 
 MYSQLD_SOCK_FILE = "/var/run/mysqld/mysqld.sock"
 MYSQLSH_SCRIPT_FILE = "/tmp/script.py"
+MYSQLD_CONFIG_PATH = "/etc/mysql/conf.d/z-custom-mysqld.cnf"
```
shayancanonical, line 18: is this constant being used?

paulomach: Not anymore! good catch

src/charm.py
```diff
@@ -4,27 +4,289 @@
...
+                # Configure instance as a cluster node
+                self._mysql.configure_instance()
+
+            except (
+                MySQLConfigureInstanceError,
+                MySQLConfigureMySQLUsersError,
+                MySQLInitialiseMySQLDError,
+                MySQLPatchDNSSearchesError,
+            ) as e:
+                self.unit.status = BlockedStatus("Unable to configure instance")
```
grobbie, line 215: Not a very helpful error message, it would be useful to give the operator insight into what they would need to do to unbork the system.

src/mysqlsh_helpers.py
```diff
@@ -241,13 +311,16 @@ def _wait_until_mysql_connection(self) -> None:
...
 
         Raises ExecError if the script gets a non-zero return code.
 
         Args:
             script: mysql-shell python script string
+            verbose: mysqlsh verbosity level
+        Returns:
+            stdout of the script
         """
         self.container.push(path=MYSQLSH_SCRIPT_FILE, source=script)
```
grobbie, line 325: there is a way to do this with pebble's `exec` method which is a lot cleaner. This was probably written before `exec` was implemented as a feature

grobbie (commented):
nice work, just a few suggestions

src/charm.py
```diff
@@ -4,27 +4,289 @@
...
+                logger.debug("Unable to configure instance: {}".format(e))
+                return
+
+            if self.unit.is_leader():
+                try:
+                    # Create the cluster when is the leader unit
+                    self._mysql.create_cluster()
+                except MySQLCreateClusterError as e:
+                    self.unit.status = BlockedStatus("Unable to create cluster")
```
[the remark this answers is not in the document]
paulomach: Cant disagree, but bear with me. We have two options here, which are:
1. Not treat the exception and allow the charm to enter in a error state
2. Treat the exception (as implemented) and leave it blocked.

Both indicates a failures and requires user intervention. So the decision boils up to how the treatment will be done.
Canonically, a blockedstatus should be resolvable through an action, but I'm not completely sure though.
For now, the current approach is to prevent ugly non treated exceptions reach the end user.

src/charm.py
paulomach: We have a guideline to keep status messages very high level, and when needed, users can check details with `juju debug-log`

src/mysqlsh_helpers.py
```diff
@@ -86,6 +106,59 @@ def __init__(
...
+        Raises MySQLPatchDNSSearchesError if the patching fails.
+        FIXME: Fix this to use the proper k8s name resolution.
+                MySQL instance are not using `group_replication_local_address`
+                to report own address to the cluster, hence this hack.
+
+        Args:
+            app_name: name of the application
+        """
+        try:
+            resolv_file = self.container.pull("/etc/resolv.conf")
```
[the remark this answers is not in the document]
paulomach: Yes, its. It's in place (with a FIXME note) while no better solution is found.

src/charm.py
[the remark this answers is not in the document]
grobbie: Is there no way to declare a Failed status from within a charm?

src/mysqlsh_helpers.py
grobbie: Ah that's weird when I read it before it looked like it was doing something else. Probably I didn't read it carefully enough 🙂

src/charm.py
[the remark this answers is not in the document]
paulomach: Not really. From statuses declaration, `BlockedStatus` is the most severe one.

src/charm.py
grobbie: I thought `debug-log` will be going away soon.

paulomach: Don't now about that. But even if it was the case, wouldn't it be super seed by something better? (I don't think we are ditching logging all together)

.github/workflows/ci.yaml
```diff
@@ -36,5 +63,18 @@ jobs:
...
+      # These two steps (download and load image) are needed until the OCI image is published to ROCKS.
+      - name: Download image built in build oci image step
+        uses: actions/download-artifact@v2
+        with:
+          name: image
+          path: /tmp
+      - name: Load Image into microk8s
+        run: |
+          /usr/bin/sg microk8s -c "microk8s ctr image import /tmp/image.tar"
+      # TODO: Remove
```
shayancanonical, line 75: remove?

tests/integration/test_charm.py
```diff
@@ -36,7 +41,111 @@ async def test_build_and_deploy(ops_test: OpsTest):
...
+@pytest.mark.skip_if_deployed
+@pytest.mark.abort_on_fail
+async def test_scale_to_ha(ops_test: OpsTest):
+    """Scale the charm to HA."""
+    await ops_test.model.applications[APP_NAME].scale(len(UNIT_IDS))
+
+    # Blocks until all units are in active state
+    logger.info("Wait for app active status")
+    sleep(10)
+    await ops_test.model.wait_for_idle(
```
shayancanonical, line 57: suggestion from raul on my PR: could do `await ops_test.model.block_until(lambda: len(ops_test.model.applications[APP_NAME].units) == len(UNIT_IDS))`

src/charm.py
jnsgruk: Debug log isn't going anywhere as such; you're on track here keeping status simple and logging out the detail.

src/mysqlsh_helpers.py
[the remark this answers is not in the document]
jnsgruk: I think we need to look more closely at this problem. This is very ugly indeed. Can you open a separate issue on the repo and we can dig in next week if you're unable to find a solution over the next couple of days. This looks like it's bound to fail at some point...

src/charm.py
```diff
@@ -4,27 +4,346 @@
...
+                return
+
+        # Add new instance to the cluster
+        try:
+            self._mysql.add_instance_to_cluster(new_instance)
+            logger.debug(f"Added instance {new_instance} to cluster")
+
+        except MySQLAddInstanceToClusterError:
+            logger.debug(f"Unable to add instance {new_instance} to cluster.")
+            event.defer()
```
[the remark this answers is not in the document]
paulomach: I'm not sure if we should block without retrying, due to some transient issues. But on the other hand it could be deferring the event forever. Not sure what's the best approach

tests/integration/test_charm.py
paulomach: Want to refactor/unify tests with machine, so not worrying to much with this right now.

.github/workflows/ci.yaml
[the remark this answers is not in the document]
paulomach: yep

src/mysqlsh_helpers.py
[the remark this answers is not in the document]
paulomach: Better solution implemented, the function was removed entirely.
Mysql required extra configuration to report it address with <hostname>.<localdomain> to it's peers.

actions.yaml
```diff
@@ -1,2 +1,11 @@
-# Copyright 2021 Canonical Ltd.
+# Copyright 2022 Canonical Ltd.
 # See LICENSE file for licensing details.
+
+get-cluster-admin-credentials:
+  description: Get the credentials for the cluster admin user
+get-server-config-credentials:
+  description: Get the credentials for the server config user
+get-root-credentials:
```
shayancanonical, line 8: i dont see this action being used anywhere. should we may be remove it? the human operator would may be only need the serverconfig and clusteradmin credentials

src/mysqlsh_helpers.py
```diff
@@ -15,18 +15,23 @@
 
 MYSQLD_SOCK_FILE = "/var/run/mysqld/mysqld.sock"
 MYSQLSH_SCRIPT_FILE = "/tmp/script.py"
+MYSQLD_CONFIG_FILE = "/etc/mysql/conf.d/z-custom.cnf"
```
shayancanonical, line 18: thoughts on renaming `/etc/mysql/conf.d/z-custom.cnf` -> `/etc/mysql/conf.d/z-report-host-custom.cnf` (or something similar)?

tests/unit/test_mysqlsh_helpers.py
```diff
@@ -161,4 +164,114 @@ def test_add_instance_to_cluster_exception(self, _run_mysqlsh_script):
         )
 
         with self.assertRaises(MySQLAddInstanceToClusterError):
-            self.mysql.add_instance_to_cluster("127.0.0.2")
+            self.mysql.add_instance_to_cluster("mysql-k8s-43.mysql-k8s-endpoints")
+
+    @patch("ops.pebble.ExecProcess")
+    @patch("ops.model.Container")
+    def test_bootstrap_instance(self, _container, _process):
```
shayancanonical, line 171: test_bootstrap_instance -> test_initialize_mysql?

tests/unit/test_mysqlsh_helpers.py
```diff
@@ -161,4 +164,114 @@ def test_add_instance_to_cluster_exception(self, _run_mysqlsh_script):
...
+        self.mysql.initialise_mysqld()
+
+        _container.exec.assert_called_once_with(
+            command=["mysqld", "--initialize-insecure", "-u", "mysql"]
+        )
+
+        _process.wait_output.assert_called_once()
+
+    @patch("ops.model.Container")
+    def test_bootstrap_instance_exception(self, _container):
```
shayancanonical, line 185: test_bootstrap_instance_exception -> test_initialize_mysqld_exception?

shayancanonical (approved):
Some minor comments, but looks good!

src/mysqlsh_helpers.py
paulomach: That's very good idea, given the propensity to create other specialized

tests/unit/test_mysqlsh_helpers.py
paulomach: done 2b62adf

tests/unit/test_mysqlsh_helpers.py
paulomach: done 2b62adf

src/mysqlsh_helpers.py
```diff
@@ -275,18 +373,92 @@ def _run_mysqlcli_script(self, script: str, password=None) -> None:
...
+
+        self._run_mysqlcli_script(
+            " ".join(commands), self.cluster_admin_password, self.cluster_admin_user
+        )
+
+    def create_custom_config_file(self, report_host: str) -> None:
+        """Create custom configuration file.
+
+        Necessary for k8s deployments.
+        Raises ExecError if the script gets a non-zero return code.
```
marceloneppel, line 438: Change to `MySQLCreateCustomConfigFileError` (the error that is being raised).

marceloneppel (approved):
LGTM! I left only two minor comments.

tests/integration/test_charm.py
```diff
@@ -36,7 +41,111 @@ async def test_build_and_deploy(ops_test: OpsTest):
...
+        result = cursor.fetchall()
+        assert result[0][0] == "hello"
+
+
+async def get_password(ops_test: OpsTest, password_key: str) -> str:
+    """Get password using the action."""
+    unit = ops_test.model.units.get(f"{APP_NAME}/0")
+    action = await unit.run_action("get-root-credentials")
+    result = await action.wait()
+    return result.results[password_key]
```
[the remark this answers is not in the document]
paulomach: Yes, this was already changed for the next PR (there are actions for each of the created users, as done in the vm charm).

src/mysqlsh_helpers.py
paulomach: Hard catch that one, fixed on 79b9f97

PR #79, opened 2022-07-13
Implement the `mysql` legacy relation

shayancanonical:
# Issue
The OSM and Kubeflow teams use the `mariadb` charm, but would like to use our charm instead. We have not yet implemented support for this legacy relation.

# Solution
Add support for the legacy relation (`mysql` interface)

# Context
Below is a snapshot of the relation data when `mariadb` is deployed with `osm-keystone`:

```
┏━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ relation (id: 0) ┃ keystone                               ┃ mariadb-k8s                            ┃
┡━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│ relation name    │ db                                     │ mysql                                  │
│ interface        │ mysql                                  │ mysql                                  │
│ leader unit      │ 0                                      │ 0                                      │
├──────────────────┼────────────────────────────────────────┼────────────────────────────────────────┤
│ application data │ ╭────────────────────────────────────╮ │ ╭────────────────────────────────────╮ │
│                  │ │ <empty>                            │ │ │ <empty>                            │ │
│                  │ ╰────────────────────────────────────╯ │ ╰────────────────────────────────────╯ │
│ unit data        │ ╭─ keystone/0* ─╮                      │ ╭─ mariadb-k8s/0* ────────────────╮    │
│                  │ │ <empty>       │                      │ │                                 │    │
│                  │ ╰───────────────╯                      │ │  database       database        │    │
│                  │                                        │ │  host           10.152.183.187  │    │
│                  │                                        │ │  password       password        │    │
│                  │                                        │ │  port           3306            │    │
│                  │                                        │ │  root_password  root            │    │
│                  │                                        │ │  user           mysql           │    │
│                  │                                        │ ╰─────────────────────────────────╯    │
└──────────────────┴────────────────────────────────────────┴────────────────────────────────────────┘

```

# Testing
Integration tests to follow in a separate PR

# Release Notes
Implement the `mysql` legacy relation


config.yaml
```diff
@@ -5,3 +5,12 @@ options:
   cluster-name:
     description: "Optional - Name of the MySQL InnoDB cluster"
     type: "string"
+# Config options for the legacy 'mysql relation'
+  user:
+    default: "mysql"
+    description: "The database username"
```
paulomach, line 11: I think its useful to describe that this is only used for legacy and it will be deprecated, and the same on L15

lib/charms/mysql/v0/mysql.py
```diff
@@ -83,7 +83,7 @@ def wait_until_mysql_connection(self) -> None:
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 1
+LIBPATCH = 2
```
paulomach, line 86: if [mysql vm pr#20](https://github.com/canonical/mysql-operator/pull/20) lands before this, the lib should be bumped

paulomach (approved):
nice and clean, lgtm - missed some unit test for `relations/mysql.py`.

config.yaml
```diff
@@ -5,3 +5,12 @@ options:
   cluster-name:
     description: "Optional - Name of the MySQL InnoDB cluster"
     type: "string"
+# Config options for the legacy 'mysql relation'
+  user:
```
jnsgruk, line 9: We could consider naming these to reflect that their purpose is for a relation. Something like `mysql-interface-user` and `mysql-interface-database`?

shayancanonical: I like the idea of making the config name reflect the legacy relation interface. However, doing so will require consumer applications to make a code change (updating the config key [here](https://github.com/canonical/bundle-kubeflow/blob/master/releases/1.4/stable/kubeflow-lite/bundle.yaml#L60)) before being able to replace the `mariadb-k8s` charm for this charm

jnsgruk: Hmm. Interesting - @DomFleischmann how disruptive would this be?

DomFleischmann: I think we might as well update to the new relation. It would still be interesting to see how this behaves during an update. What I expect us to do would basically be the following:

1. Deploy the new mysql charm alongside the new one.
2.  Upgrade the charms that consume it, for that release they will suppport both relations. 
3. Migrate the data to the new mysql
4. Relate the new mysql with the new charms
5. Delete the old relation and remove the old charms.

Does that sound like the right way of migrating to the new charms @shayancanonical ?

DomFleischmann: > Do we have long lasting kubeflow deployments?

@shayancanonical Yes, and long lasting customer deployments by the time we move to the new charm.

> I believe that with this PR, one would be able to swap out mariadb-k8s with mysql-k8s without any difference (given that all tables have primary keys). This is what we call the legacy relation (support for which is introduced in this PR).

In that case if we change the config name it will only affect the new bundle with the new charm, this is not a big deal and can be done without an issue. 

The upgrade/migration path is something that probably requires more discussion in another forum.

config.yaml
shayancanonical: Addressed in 16a0a25a5b7c82698b3fa16c84266763cb9a4193

config.yaml
shayancanonical: Changed config keys to include `mysql-interface-` prefix in 16a0a25a5b7c82698b3fa16c84266763cb9a4193

marceloneppel (approved):
LGTM!

PR #80, opened 2022-07-13
Implement integration tests for mysql legacy relation

shayancanonical:
# Issue
1. There exist no integration tests for the legacy `mysql` relation interface with `osm-keystone` charm
2. There exist no integration tests for the legacy `mysql` relation interface with `kfp-api` charm
3. The integration tests are not parallelized

# Solution
1 and 2. Implement the integration tests
3. Parallelize the integration tests by running them in different github action workflows

# Release Notes
Implement integration tests for `mysql` relation


.github/workflows/ci.yaml
```diff
@@ -40,4 +40,19 @@ jobs:
           channel: 1.23/stable
           bootstrap-options: "--agent-version 2.9.29"
       - name: Run integration tests
-        run: tox -e integration
+        run: tox -e integration-ha
+  integration-legacy-mysql-test-microk8s:
```
MiaAltieri, line 44: nit: integration-legacy-relations sounds better, I am not sure if `-mysql-test-microk8s` adds useful info

tests/integration/test_charm.py
```diff
@@ -127,6 +129,7 @@ async def test_consistent_data_replication_across_cluster(ops_test: OpsTest) ->
 
 @pytest.mark.order(3)
 @pytest.mark.abort_on_fail
+@pytest.mark.ha_tests
```
MiaAltieri, line 132: The tests for charms are going to grow quite large, it might be useful to split things into separate files instead of using `@pytest.mark.ha_tests`. Check out [this WIP](https://github.com/canonical/mongodb-operator/pull/61) PR that does this. But if you prefer your way of doing this with `@pytest.mark.ha_tests` then feel free to keep as is :)

tests/integration/test_legacy_mysql.py
```diff
@@ -0,0 +1,208 @@
...
+            wait_for_exact_units=1,
+        )
+
+        # Relate the mysql charm with the osm-keystone charm
+        await ops_test.model.relate(f"{OSM_KEYSTONE_APP_NAME}:db", f"{DATABASE_APP_NAME}:mysql")
+        await ops_test.model.wait_for_idle(
+            apps=[OSM_KEYSTONE_APP_NAME, DATABASE_APP_NAME],
+            status="active",
+            raise_on_blocked=False,
+            timeout=1000,
```
MiaAltieri, line 105: Remember that "Not waiting for relations can lead to race-conditions in integration tests", see [this](https://github.com/canonical/mongodb-operator/blob/d49bac6436f7d676be4a4ef6aebf727a245ee815/tests/integration/relation_tests/legacy_relations/test_charm_legacy_relations.py#L186-L201) or Paulo's use of it (I can't find his code for this, maybe you know where it is?). I think Paulo's is nicer since he uses `block_until`. If you don't need to wait, then include a comment as to why [see here for an example](https://github.com/canonical/mongodb-operator/blob/d49bac6436f7d676be4a4ef6aebf727a245ee815/tests/integration/relation_tests/legacy_relations/test_charm_legacy_relations.py#L45)

tests/integration/test_legacy_mysql.py
```diff
@@ -0,0 +1,208 @@
...
+        )
+
+        # Ensure that the keystone database exists
+        show_databases_sql = [
+            "SHOW DATABASES",
+        ]
+
+        random_unit = ops_test.model.applications[DATABASE_APP_NAME].units[0]
```
MiaAltieri, line 113: not sure if this unit is actually random
```suggestion
        db_unit = ops_test.model.applications[DATABASE_APP_NAME].units[0]
```

tests/integration/test_legacy_mysql.py
```diff
@@ -0,0 +1,208 @@
...
+            )
+            assert "keystone" in output
+
+        # Scale down the keystone application
+        await scale_application(ops_test, OSM_KEYSTONE_APP_NAME, 0)
+
+        await ops_test.model.remove_application(OSM_KEYSTONE_APP_NAME)
+
+
+# TODO: uncomment when ready to integrate with kfp-api
```
MiaAltieri, line 133: I'll review this once its uncommented, perhaps remove it for now if its not being used? :)

tests/integration/test_legacy_mysql.py
```diff
@@ -0,0 +1,208 @@
...
+        assert len(ops_test.model.applications[DATABASE_APP_NAME].units) == 3
+        for unit in ops_test.model.applications[DATABASE_APP_NAME].units:
+            assert unit.workload_status == "active"
+
+
+@pytest.mark.order(2)
+@pytest.mark.abort_on_fail
+@pytest.mark.legacy_mysql_tests
+async def test_osm_keystone_bundle_mysql(ops_test: OpsTest) -> None:
+    """Deploy the osm keystone bundle to test the legacy 'mysql' relation.
```
MiaAltieri, line 71: this function is confusing for a noob like me, can you be more specific about what is exactly getting tested by the osm keystone bundle here?

tests/integration/test_legacy_mysql.py
```diff
@@ -0,0 +1,208 @@
...
+                show_databases_sql,
+            )
+            assert "keystone" in output
+
+        # Scale down the keystone application
+        await scale_application(ops_test, OSM_KEYSTONE_APP_NAME, 0)
+
+        await ops_test.model.remove_application(OSM_KEYSTONE_APP_NAME)
+
+
```
MiaAltieri, line 132: it would be nice to add a CRUD test for the keystone app

MiaAltieri (changes requested):

tests/integration/test_legacy_mysql.py
paulomach: I'll say to keep it until the issue in the upstream is solved.
@shayancanonical , in such case, should you keep the `dontmerge` tag? (Also for keeping in mind the grafana issue)

tests/integration/test_charm.py
paulomach: To clarify, they are mostly in segregated files already.
This is used in the tox environment for each test, so it skips all other `marks`, avoiding the necessity for writing the test file path.

paulomach (commented):
Wouldn't be the case to keep this as `Draft` while we wait for the changes in the upstreams?

tests/integration/test_legacy_mysql.py
zmraul: You can also use `block_until` with timeout and a retry, very similar to what we would do with a tenacity retry:
```python

# wait on test
await ops_test.model.block_until(
    lambda: check_application_status(ops_test, "discourse-k8s") == "active",
    timeout=600,
    wait_period=5,
)

def check_application_status(ops_test: OpsTest, app_name: str) -> str:
    """Return the application status for an app name."""
    model_name = ops_test.model.info.name
    proc = subprocess.check_output(f"juju status --model={model_name}".split())
    proc = proc.decode("utf-8")

    statuses = {"active", "maintenance", "waiting", "blocked"}
    for line in proc.splitlines():
        parts = line.split()
        # first line with app name will have application status
        if parts and parts[0] == app_name:
            # NOTE: intersects possible statuses with the list of values:
            # this is done because sometimes version number exists and
            # sometimes it doesn't on juju status output
            find_status = list(statuses & set(parts))
            if find_status:
                return find_status[0]

```

shayancanonical:
Closing as this PR is really outdated. We can refer back to it when the kubeflow team adopts kfp changes in a future version.

PR #81, opened 2022-07-15
Implement scale down without relying on teardown events

shayancanonical:
# Issue
There is a [bug](https://bugs.launchpad.net/juju/+bug/1951415) in juju that is preventing teardown event from firing in k8s. Therefore, we have to implement scale down of instances without relying on these teardown events.

# Solution
We rely on `relation_departed` events where (only on the juju leader unit) we remove unreachable instances from the cluster. If the previous juju leader unit was scaled down, we would remove unreachable instances in the `leader-elected` event handler.

# Context
We can remove unreachable instances by using `mysqlsh cluster.remove_instance` with options `force=true`
If there are 2 instances where 1 instance is unreachable, the cluster has lost quorum. In such a scenario, we can use `mysqlsh cluster.force_quorum_using_partition_of()` to restore quorum using an online node.

# Release Notes
Implement scale down without relying on teardown events


src/charm.py
```diff
@@ -349,6 +395,33 @@ def _on_peer_relation_changed(self, event: RelationChangedEvent) -> None:
             self.unit.status = ActiveStatus()
             logger.debug(f"Instance {instance_label} is cluster member")
 
+    def _on_peer_relation_departed(self, event: RelationDepartedEvent) -> None:
```
paulomach, line 398: Remind me here, but do we need to update ipAllowList on every node?

tox.ini
```diff
@@ -68,10 +68,11 @@ commands =
 [testenv:integration]
 description = Run integration tests
 deps =
-    pytest
     juju
-    pytest-operator
     mysql-connector-python
+    pytest
+    pytest-operator
+    pytest-order
```
paulomach, line 75: you might need to add [this from marcelo PR](https://github.com/canonical/postgresql-k8s-operator/pull/20/files) here also

paulomach (commented):
Some minor comments here and there. Happy with the improvement in itests!

tests/integration/test_charm.py
```diff
@@ -4,148 +4,158 @@
...
-    cnx = db_connect(host_ip, root_password)
-    cursor = cnx.cursor()
-    # ensure table cleanup (when running with --no-deploy)
-    # create a table and insert a row
-    queries = [
-        "DROP TABLE IF EXISTS mysql.charmtest;",
-        "CREATE TABLE mysql.charmtest (test_field VARCHAR(255) PRIMARY KEY);",
-        "INSERT INTO mysql.charmtest VALUES ('hello');",
+    # Allow time for the data to be replicated
+    time.sleep(2)
```
marceloneppel, line 106: Maybe you can change it to something like below to enclose the `execute_queries_on_unit` calls:

```python
try:
    for attempt in Retrying(stop=stop_after_delay(10), wait=wait_fixed(3)):
        with attempt:
            # Execute the logic that can fail until some time has passed.
except RetryError:
    return False
```

marceloneppel (approved):
Great work Shayan!

tests/integration/test_charm.py
```diff
@@ -4,148 +4,158 @@
...
 logger = logging.getLogger(__name__)
 
 METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
 APP_NAME = METADATA["name"]
+CLUSTER_NAME = "test_cluster"
 
 UNIT_IDS = [0, 1, 2]
 
 
+@pytest.mark.order(1)
```
WRFitch, line 33: Big fan of these

src/charm.py
```diff
@@ -198,6 +214,36 @@ def _on_leader_elected(self, _):
...
+        if self._is_peer_data_set and self._is_cluster_initialized:
+            self.unit.status = MaintenanceStatus("Removing unreachable instances")
+
+            # Remove unreachable instances from the cluster
+            self._mysql.remove_instances_not_online()
+
+            # Since we do not have name of the prior leader unit, use planned units
+            # to remove units with index >= planned units from the allowlist
+            planned_units = self.app.planned_units()
+            allowlist = set(self._peers.data[self.app].get("allowlist").split(","))
```
MiaAltieri, line 227: sorry noob to this lingo, whats an "allow list"

src/charm.py
```diff
@@ -198,6 +214,36 @@ def _on_leader_elected(self, _):
...
+            new_allowlist = set(allowlist)
+
+            for fqdn in allowlist:
```
MiaAltieri, line 230: you might be able to simplify this with some set math like how Mykola did when he handled this issue in the [k8s charm](https://github.com/delgod/mongodb-k8s-operator/blob/f7e8690b7e7552e226cc57ccf51ba24b562243af/src/charm.py#L177-L180) it might simplify the logic, but it would require that you keep a list of the current set members. 🤷

WRFitch (approved):
LGTM, I did have a question about blocking if we fail to remove instances though

src/charm.py
```diff
@@ -349,6 +395,33 @@ def _on_peer_relation_changed(self, event: RelationChangedEvent) -> None:
             self.unit.status = ActiveStatus()
             logger.debug(f"Instance {instance_label} is cluster member")
 
+    def _on_peer_relation_departed(self, event: RelationDepartedEvent) -> None:
+        """Handle the relation departed event.
+
+        Update the allowlist to remove the departing unit for all units.
```
MiaAltieri, line 401: curious why does this need to be done for all units? What was the design motivation on this? Was it SQL based?

If its SQL based It would be nice to have a description on that here :)

tests/unit/test_mysqlsh_helpers.py
```diff
@@ -148,6 +149,112 @@ def test_configure_mysql_users_exception(self, _run_mysqlcli_script):
...
+        # disable tenacity retry
+        self.mysql.remove_instances_not_online.retry.retry = tenacity.retry_if_not_result(
+            lambda x: True
+        )
+
+        _get_cluster_status.return_value = None
+        with self.assertRaises(MySQLRemoveInstancesNotOnlineRetryError):
+            self.mysql.remove_instances_not_online()
+
+        get_cluster_status_return = {
```
MiaAltieri, line 233: might be worth storing in a file and re-loading

tests/unit/test_mysqlsh_helpers.py
```diff
@@ -148,6 +149,112 @@ def test_configure_mysql_users_exception(self, _run_mysqlcli_script):
         with self.assertRaises(MySQLConfigureMySQLUsersError):
             self.mysql.configure_mysql_users()
 
+    @patch("mysqlsh_helpers.MySQL.get_cluster_status")
+    @patch("mysqlsh_helpers.MySQL._run_mysqlsh_script")
+    def test_remove_instances_not_online(self, _run_mysqlsh_script, _get_cluster_status):
+        """Test a successful execution of remove_instances_not_online."""
+        get_cluster_status_return = {
```
MiaAltieri, line 156: might be worth storing this as a file and reading in later

tests/integration/test_charm.py
```diff
@@ -4,148 +4,158 @@
...
+
+        cluster_status = await get_cluster_status(ops_test, random_unit)
+        online_member_addresses = [
+            member["address"]
+            for _, member in cluster_status["defaultreplicaset"]["topology"].items()
+            if member["status"] == "online"
+        ]
+        assert len(online_member_addresses) == 5
+
+        await scale_application(ops_test, APP_NAME, 1)
```
MiaAltieri, line 142: noob question, doesn't this not always tear down the primary, ie the priamry could have been the single unit remaining? (ie primary removal will get tested 4/5 of the time?)

MiaAltieri (approved):

src/mysqlsh_helpers.py
```diff
@@ -241,6 +253,77 @@ def update_allowlist(self, allowlist: str) -> None:
...
+                if there was an issue removing not online instances
+        """
+        try:
+            cluster_status = self.get_cluster_status()
+            if not cluster_status:
+                raise MySQLRemoveInstancesNotOnlineRetryError("Unable to retrieve cluster status")
+
+            # If the cluster has no quorum, force quorum using partition of
+            # the first online instance
+            if cluster_status["defaultreplicaset"]["status"] == "no_quorum":
```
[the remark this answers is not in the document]
shayancanonical: yep, they're lowercase-d [in the method](https://github.com/canonical/mysql-k8s-operator/blob/main/lib/charms/mysql/v0/mysql.py#L594) that returns them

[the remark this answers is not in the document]
paulomach: doh! I've wrote that..

src/mysqlsh_helpers.py
```diff
@@ -241,6 +253,77 @@ def update_allowlist(self, allowlist: str) -> None:
...
+                force_quorum_commands = (
+                    f"shell.connect('{self.cluster_admin_user}:{self.cluster_admin_password}@{self.instance_address}')",
+                    f"cluster = dba.get_cluster('{self.cluster_name}')",
+                    f"cluster.force_quorum_using_partition_of('{self.cluster_admin_user}@{online_member_address}', '{self.cluster_admin_password}')",
+                )
+
+                self._run_mysqlsh_script("\n".join(force_quorum_commands))
+
+                # Refresh the cluster status after forcing quorum
+                cluster_status = self.get_cluster_status()
```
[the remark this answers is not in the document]
shayancanonical: Moved to a new private function in a9bd5bb1a336943a9c0d601492f594aaddd073e8

src/charm.py
```diff
@@ -349,6 +395,33 @@ def _on_peer_relation_changed(self, event: RelationChangedEvent) -> None:
...
+            event.defer()
+            return
+
+        if not self.unit.is_leader():
+            return
+
+        self._peers.data[self.app]["allowlist"] = ",".join(allowlist)
+
+        self.unit.status = MaintenanceStatus("Removing unreachable instances")
+        self._mysql.remove_instances_not_online()
```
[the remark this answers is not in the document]
shayancanonical: Good point! Added in a9bd5bb1a336943a9c0d601492f594aaddd073e8

tests/integration/test_charm.py
shayancanonical: Added in a9bd5bb1a336943a9c0d601492f594aaddd073e8

src/charm.py
```diff
@@ -198,6 +214,36 @@ def _on_leader_elected(self, _):
...
+
+            for fqdn in allowlist:
+                unit_number = int(fqdn.split(".")[0].split("-")[1])
+
+                if unit_number >= planned_units:
+                    new_allowlist.remove(fqdn)
+
+            try:
+                self._mysql.update_allowlist(",".join(new_allowlist))
+            except MySQLUpdateAllowListError:
```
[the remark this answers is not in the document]
shayancanonical: Great point! Added a try/catch and move into `BlockedStatus` in a9bd5bb1a336943a9c0d601492f594aaddd073e8

tox.ini
```diff
@@ -68,10 +68,11 @@ commands =
...
     -r{toxinidir}/requirements.txt
 commands =
     pytest -v --tb native --ignore={[vars]tst_path}unit --log-cli-level=INFO -s {posargs}
```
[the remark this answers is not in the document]
shayancanonical: Added in a9bd5bb1a336943a9c0d601492f594aaddd073e8

src/charm.py
shayancanonical: We're only setting it on the leader unit, but I am not sure if it needs to be set on every node. Will get Mykola's direction and make any changes in a followup. For now, the ip allow list is only updated on the leader node

src/charm.py
shayancanonical: An [allowlist](https://dev.mysql.com/doc/refman/8.0/en/group-replication-options.html#sysvar_group_replication_ip_allowlist) is a list of hostnames/IPs that are permitted to be part of the group (cluster) in MySQL

src/charm.py
shayancanonical: We actually may be handling allowlists incorrectly (we're only setting them on the juju leader unit, whereas we might need to set them in all units). Will followup after speaking with Mykola

src/charm.py
shayancanonical: I thought that it needs to be done in all units since we need to make sure no other unit can join the group. However, I will speak with Mykola about how we need allowlists to work, and make any changes in a followup

tests/unit/test_mysqlsh_helpers.py
shayancanonical: moving to a file may be overkill for now, but moved it to a constant at the top of file in a9bd5bb1a336943a9c0d601492f594aaddd073e8 (to be reused)

tests/unit/test_mysqlsh_helpers.py
shayancanonical: moving to a file may be overkill for now, but moved it to a constant at the top of file in a9bd5bb1a336943a9c0d601492f594aaddd073e8 (to be reused)

tests/integration/test_charm.py
shayancanonical: the integration test is not actually testing for the scale down of the primary - instead testing how the loss of quorum is dealt with by the charm when scaling down (when going from 5 online nodes to 1 online node)

PR #84, opened 2022-07-20
Enable nodes with IPV6 addresses in the group replication 

shayancanonical:
# Issue
We are running into this [issue/error trace](https://github.com/canonical/mysql-k8s-operator/runs/7434800404?check_suite_focus=true#step:4:6640) when running integration tests on github actions.

[Further research](https://dev.mysql.com/doc/refman/8.0/en/group-replication-options.html#sysvar_group_replication_ip_allowlist) shows that connections in the group may be refused if the protocol is different that the one in the `group_replication_ip_allowlist`. Since we are setting an ipv4 cidr (`0.0.0.0/0`), connections from ipv6 addresses are being refused.

# Solution
Set the `group_replication_ip_allowlist` as `0.0.0.0/0,::/0` to allow any ipv4 or ipv6 address.

# Release Notes
Enable nodes with IPV6 addresses in the group replication

### DIFF
--- a/src/mysqlsh_helpers.py
+++ b/src/mysqlsh_helpers.py
@@ -216,6 +216,7 @@ class MySQL(MySQLBase):
         commands = (
             "INSTALL PLUGIN group_replication SONAME 'group_replication.so'",
             f"SET PERSIST group_replication_local_address='{self.instance_address}:33061'",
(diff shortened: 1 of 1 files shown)


MiaAltieri (approved):
excellent PR description

paulomach (approved):
LGTM

PR #86, opened 2022-07-22
Update mysql lib to v0.4 to pull in fixes

shayancanonical:
# Issue
We have a few fixes in the shared mysql lib (v0.4) from the mysql-operator repo that we have not yet fixed into this repository.

# Solution
Update the lib to v0.4

# Release Notes
n/a

### DIFF
--- a/lib/charms/mysql/v0/mysql.py
+++ b/lib/charms/mysql/v0/mysql.py
@@ -69,7 +69,7 @@ import json
 import logging
 import re
 from abc import ABC, abstractmethod
-from typing import List, Tuple
+from typing import Iterable, List, Optional, Tuple
 
 from tenacity import retry, retry_if_exception_type, stop_after_attempt, wait_random
 
@@ -83,7 +83,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 2
+LIBPATCH = 4
 
 UNIT_TEARDOWN_LOCKNAME = "unit-teardown"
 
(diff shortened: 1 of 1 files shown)


paulomach (approved):

lib/charms/mysql/v0/mysql.py
```diff
@@ -825,13 +857,66 @@ def get_cluster_primary_address(self, connect_instance_address: str = None) -> s
...
+        """Get the addresses of the cluster's members.
+
+        Returns:
+            Iterable of members addresses
+        """
+        get_cluster_members_commands = (
+            f"shell.connect('{self.cluster_admin_user}:{self.cluster_admin_password}@{self.instance_address}')",
+            f"cluster = dba.get_cluster('{self.cluster_name}')",
+            "members = ','.join((member['address'] for member in cluster.describe()['defaultReplicaSet']['topology']))",
+            "print(f'<MEMBERS>{members}</MEMBERS>')",
```
marcoppenheimer, line 870: Perhaps I'm not understanding MySQL, but I'm confused as to what's going on here.
Are we exec-ing generated Python code against `/usr/bin/mysqlsh`?

If so that seems quite dangerous and REALLY hard to debug.

Is something stopping us from doing the same as a subprocess using returned data from MySQL API to get the values we want (I think 'address','members' etc)?

marcoppenheimer: Feel free to ignore this if this is totally fine and safe. Just looks a bit wild 😄

marcoppenheimer (approved):
Approving because I reckon my comment isn't a blocker, but would love to know more on it!

lib/charms/mysql/v0/mysql.py
shayancanonical: The thing is, `mysqlsh` (which is an advanced command line tool) has an extensive (python) [API](https://dev.mysql.com/doc/dev/mysqlsh-api-python/8.0/annotated.html) to manage the cluster and get all kinds of metadata on the cluster. If we were to use a mysql client, these utilities wouldn't be available to us. `mysqlsh` has the ability to interpret and execute python code, which is what we're doing in a subprocess (by calling `_run_mysqlsh_script` -> we write the python code to a file and then pass the file to `mysqlsh` to execute)

PR #88, opened 2022-08-10
Add support for the osm-mysql interface

shayancanonical:
# Issue
The `osm` bundle uses the `osm-mariadb` charm, but would like to use this charm instead. However, the `mysql` relation interface (used by kubeflow) creates a restricted database user for use by the application. The `osm` bundle requires users with escalated privileges as it has an application (`osm-pol`) that creates a user and grants it privileges.

# Solution
Implement a separate `osm-mysql` relation interface, where the root and application user have escalated privileges to be able to create users and provide them with grants.

# Release Notes
Add support for the osm-mysql interface


tox.ini
```diff
@@ -68,11 +68,35 @@ commands =
...
+description = Run integration tests
+deps =
+    juju==2.9.11 # juju 3.3.0 has issues with retrieving action results
+    mysql-connector-python
+    pytest
+    pytest-operator
+    pytest-order
+    -r{toxinidir}/requirements.txt
+commands =
+    pytest -v --tb native --ignore={[vars]tst_path}unit --log-cli-level=INFO -s {posargs} --asyncio-mode=auto -m "ha_tests"
```
marceloneppel, line 90: Paulo had a [good suggestion](https://github.com/canonical/postgresql-operator/pull/10#discussion_r924885319) about moving the asyncio mode to `pyproject.toml` so it's possible to avoid specifying it here in each environment.

src/mysqlsh_helpers.py
```diff
@@ -340,6 +356,129 @@ def remove_instances_not_online(self) -> None:
...
+
+    def delete_users_with_label(self, label_name: str, label_value: str) -> None:
+        """Delete users for a unit.
+
+        Args:
+            unit_name: The name of the unit for which to delete mysql users for
+
+        Raises:
+            MySQLDeleteUsersForUnitError if there is an error deleting users for the unit
+        """
```
marceloneppel, line 446: The docstring needs to be updated.

marceloneppel (approved):
LGTM! Great job, Shayan!

src/relations/mysql.py
```diff
@@ -142,6 +142,9 @@ def _on_mysql_relation_broken(self, event: RelationBrokenEvent) -> None:
         if not self.charm.unit.is_leader():
             return
 
+        if len(self.charm.model.relations[LEGACY_MYSQL]) > 1:
```
MiaAltieri, line 145: add a comment for why you're doing this

src/relations/osm_mysql.py
```diff
@@ -0,0 +1,153 @@
...
+
+        self.framework.observe(self.charm.on.leader_elected, self._on_leader_elected)
+        self.framework.observe(
+            self.charm.on[LEGACY_OSM_MYSQL].relation_created, self._on_osm_mysql_relation_created
+        )
+        self.framework.observe(
+            self.charm.on[LEGACY_OSM_MYSQL].relation_broken, self._on_osm_mysql_relation_broken
+        )
+
+    def _get_or_set_password_in_peer_databag(self, username: str) -> str:
```
MiaAltieri, line 35: `_get_password_in_peer_databag `

src/relations/osm_mysql.py
```diff
@@ -0,0 +1,153 @@
...
+        password = generate_random_password(PASSWORD_LENGTH)
+        peer_databag[f"{username}_password"] = password
+
+        return password
+
+    def _on_leader_elected(self, _) -> None:
+        """Handle the leader elected event.
+
+        Retrieves relation data from the peer relation databag and copies
+        the relation data into the new leader unit's databag.
```
MiaAltieri, line 58: Question: why store everything in the leaders databag rather than in the app data bag? Then you might not need t o copy things over all the time?

src/relations/osm_mysql.py
```diff
@@ -0,0 +1,153 @@
...
+        being elected).
+        """
+        if not self.charm.unit.is_leader():
+            return
+
+        # Wait until on-config-changed event is executed
+        # (wait for root password to have been set)
+        if not self.charm._is_peer_data_set:
+            event.defer()
+            return
```
MiaAltieri, line 95: In order `self.charm._is_peer_data_set` the  ` on-config-changed ` event needs to run deferring here can cause this event to continuously defer making it take much longer for the config-changed hook to get on the queue before this event re-queues. This can lead to downtime/poor performance. Instead I suggest you set a flag and trigger and event that will re-call `_on_osm_mysql_relation_created`.

Mykola did something similar in his k8s charm [here](https://github.com/delgod/mongodb-k8s-operator/blob/ad178a7e2406f3b805ac73d5dfb7927b123c6deb/lib/charms/mongodb_libs/v0/mongodb_provider.py#L60-L63) and instead of re-triggering the event, he creating a function for himself to call once his condition was met by the start hook, see [here](https://github.com/delgod/mongodb-k8s-operator/blob/ad178a7e2406f3b805ac73d5dfb7927b123c6deb/src/charm.py#L142)

I think Paulo mentioned that this delay will only occur in a bundle but I think its worth looking out for.

src/relations/osm_mysql.py
```diff
@@ -0,0 +1,153 @@
...
+
+        logger.warning("DEPRECATION WARNING - `osm-mysql` is a legacy interface")
+
+        username = self.charm.config.get("osm-mysql-interface-user")
+        database = self.charm.config.get("osm-mysql-interface-database")
+
+        if not username or not database:
```
MiaAltieri, line 102: comment for why

src/relations/osm_mysql.py
```diff
@@ -0,0 +1,153 @@
...
+    def _on_osm_mysql_relation_broken(self, event: RelationBrokenEvent) -> None:
+        """Handle the 'mysql' legacy relation broken event.
+
+        Delete the application user created in the relation created
+        event handler.
+        """
+        if not self.charm.unit.is_leader():
+            return
+
+        if len(self.charm.model.relations[LEGACY_OSM_MYSQL]) > 1:
```
MiaAltieri, line 148: why?

tests/integration/test_charm.py
```diff
@@ -127,6 +127,7 @@ async def test_consistent_data_replication_across_cluster(ops_test: OpsTest) ->
 
 @pytest.mark.order(3)
 @pytest.mark.abort_on_fail
+@pytest.mark.ha_tests
```
MiaAltieri, line 130: might be nice to have. a different file since there will be _a lot_ of HA tests, but I am not opinionated on this right now :)

tests/integration/test_osm_mysql.py
```diff
@@ -0,0 +1,161 @@
...
+                        assert "keystone" in output
+                        assert "pol" in output
+
+                        # test that osm-pol successfully creates tables
+                        output = await execute_queries_on_unit(
+                            unit_address,
+                            server_config_credentials["username"],
+                            server_config_credentials["password"],
+                            get_count_pol_tables,
+                        )
```
MiaAltieri, line 158: might be worth checking delete and update functions

tests/integration/test_osm_mysql.py
```diff
@@ -0,0 +1,161 @@
...
+                        assert output[0] > 0
+        except RetryError:
+            assert False
```
MiaAltieri, line 161: might be worth checking that it has "users with escalated privileges as it has an application (osm-pol) that creates a user and grants it privileges." since this is what makes this relation unique.

tests/integration/test_osm_mysql.py
```diff
@@ -0,0 +1,161 @@
...
+        await ops_test.model.block_until(lambda: is_relation_joined(ops_test, "mongodb", "mongo"))
+
+        await ops_test.model.relate("osm-pol:kafka", "osm-kafka:kafka")
+        await ops_test.model.block_until(lambda: is_relation_joined(ops_test, "kafka", "kafka"))
+
+        await ops_test.model.relate("osm-pol:mysql", f"{APP_NAME}:osm-mysql")
+        await ops_test.model.block_until(
+            lambda: is_relation_joined(ops_test, "mysql", "osm-mysql")
+        )
+
```
MiaAltieri, line 123: I think it would be nice to break this into two tests

`test_deploy_and_relate` and `test_osm_pol_ops`

tox.ini
```diff
@@ -68,11 +68,35 @@ commands =
...
+
+[testenv:integration-osm-mysql]
```
MiaAltieri, line 92: be sure you add this test to `.github/workflows/release.yaml` otherwise it won't run these tests before releasing to charmhub

MiaAltieri (commented):
Looking AWESOME Shayan! 🤩

src/relations/osm_mysql.py
shayancanonical: The issue is that reactive charms (which the osm charms are) read relation data from the leader unit's databag. The `leader-elected` event handler exists to make sure that if a new leader is elected, the relation data is copied from the app databag to the new leader's databag

src/relations/osm_mysql.py
shayancanonical: I'm not sure that this idea will work in this case. There are 2 conditions that must be met before the event handler (function) can be called:
1. The `on-config-changed` should have completed execution at least once
2. The `osm-mysql` relation must have been created
With these conditions and machinery at hand, I believe that the existing approach is sufficient (and I agree with Paulo that this will only happen once in a bundle at the time of relation being created)

tests/integration/test_osm_mysql.py
shayancanonical: After double-checking the application (osm-pol), the existence of the `pol` database and tables within it implies that the escalated user has been created correctly (as osm-pol requires privileges to create a user to be able to run the migrations). Additionally, it's hard to test for the existence of a user created by osm-pol as its naming scheme is not straight-forward

tests/integration/test_osm_mysql.py
shayancanonical: I'm unsure which user we should test update + delete grants for. Ideally, it would be the user created by the osm-pol, but we are unable to guess the username that the application creates. Instead, here we just check that the application can run migrations (and creates tables)

tox.ini
shayancanonical: addressed in 2f6e6e434f2bab7f2a508153eb234fb73edf2bed

src/mysqlsh_helpers.py
shayancanonical: addressed in 2f6e6e434f2bab7f2a508153eb234fb73edf2bed

src/relations/osm_mysql.py
```diff
@@ -0,0 +1,153 @@
...
+            return
+
+        logger.warning("DEPRECATION WARNING - `osm-mysql` is a legacy interface")
+
+        self.charm._mysql.delete_users_with_label("label", "osm-mysql-legacy-relation")
```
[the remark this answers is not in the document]
shayancanonical: addressed in 2f6e6e434f2bab7f2a508153eb234fb73edf2bed

src/relations/osm_mysql.py
```diff
@@ -0,0 +1,153 @@
...
+            event.relation.data[self.charm.unit].update(updates)
+
+            return
+
+        password = self._get_or_set_password_in_peer_databag(username)
+
+        self.charm._mysql.create_database(database)
+        self.charm._mysql.create_user(username, password, "osm-mysql-legacy-relation")
+        self.charm._mysql.escalate_user_privileges("root")
+        self.charm._mysql.escalate_user_privileges(username)
```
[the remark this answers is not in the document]
shayancanonical: addressed in 2f6e6e434f2bab7f2a508153eb234fb73edf2bed

src/relations/mysql.py
shayancanonical: addressed in 2f6e6e434f2bab7f2a508153eb234fb73edf2bed

src/relations/osm_mysql.py
shayancanonical: addressed in 2f6e6e434f2bab7f2a508153eb234fb73edf2bed

src/relations/osm_mysql.py
shayancanonical: addressed in 2f6e6e434f2bab7f2a508153eb234fb73edf2bed

src/relations/osm_mysql.py
shayancanonical: addressed in 2f6e6e434f2bab7f2a508153eb234fb73edf2bed

tests/integration/test_charm.py
shayancanonical: we will address this comment in a different PR when the need arises (when we have added more HA tests)

tests/integration/test_osm_mysql.py
shayancanonical: addressed in 2f6e6e434f2bab7f2a508153eb234fb73edf2bed

tox.ini
shayancanonical: addressed in 2f6e6e434f2bab7f2a508153eb234fb73edf2bed

src/mysqlsh_helpers.py
marceloneppel: It seems that it was not updated in that commit.

shayancanonical: i believe i updated it [here](https://github.com/canonical/mysql-k8s-operator/pull/88/files#diff-6e566a7b93efc3f3e7ac0e6d8989256725f876f06af78c45d0ae3815d4bcf82eR454) -> the name of the function is slightly different than the copy pasted one

marceloneppel: Ah, sorry Shayan! I haven't made it clear that more parts of the docstring need changes (the args section and maybe the initial description can be updated; the [args section](https://github.com/canonical/mysql-k8s-operator/commit/2f6e6e434f2bab7f2a508153eb234fb73edf2bed#diff-6e566a7b93efc3f3e7ac0e6d8989256725f876f06af78c45d0ae3815d4bcf82eR451) is not matching the list of arguments of this method)

shayancanonical: ah sorry, i completely missed that. corrected in f375b063af5362d0b622dce487e68f2bfd54a630

marceloneppel (approved):

src/relations/mysql.py
paulomach: `s/osm-mysql/mysql/` ?

src/relations/osm_mysql.py
```diff
@@ -0,0 +1,190 @@
...
+
+        user_exists = False
+        try:
+            user_exists = self.charm._mysql.does_mysql_user_exist(username, "%")
+        except MySQLCheckUserExistenceError:
+            self.charm.unit.status = BlockedStatus("Failed to check user existence")
+            return
+
+        # Only execute if the application user does not exist
+        if user_exists:
```
paulomach, line 130: If i understood correctly, this case will be valid when the joining unit is from a second application being related using `osm-mysql`.
right? or am I out to lunch?

tests/integration/test_osm_mysql.py
```diff
@@ -0,0 +1,166 @@
...
+        )
+        await ops_test.model.block_until(
+            lambda: len(ops_test.model.applications["osm-zookeeper"].units) == 1
+        )
+        await ops_test.model.block_until(
+            lambda: len(ops_test.model.applications["osm-mongodb"].units) == 1
+        )
+
+        await ops_test.model.relate("osm-kafka:zookeeper", "osm-zookeeper:zookeeper")
+        await ops_test.model.block_until(
```
paulomach, line 91: Consider adding `timeout`'s to all `block_until` calls.

tox.ini
```diff
@@ -68,11 +68,35 @@ commands =
 [testenv:integration]
 description = Run integration tests
 deps =
-    juju
+    juju==2.9.11 # juju 3.3.0 has issues with retrieving action results
```
paulomach, line 71: did you meant `2.9.33` ? (the current stable in 2.x series)

paulomach (approved):
Looking smooth - some minor comments to take a peak!

src/relations/osm_mysql.py
shayancanonical: you're correct. we cannot prevent multiple applications from being related on the same interface/endpoint. this check is just address this edge case

tests/integration/test_osm_mysql.py
shayancanonical: added in [750358c](https://github.com/canonical/mysql-k8s-operator/pull/88/commits/750358c2a892641c312e45f948db18bc9efe8924)

tox.ini
shayancanonical: the highest 2.9 version for python-libjuju is `2.9.11` (i tried to change it to `2.9.33`, but got an error)

PR #89, opened 2022-08-16
Port of new relation implementation to k8s

paulomach:
# Issue

 Missing "new-relations" implementation on k8s.

# Solution

This humongous PR is almost entirely copy-and-paste from MySQL machine charm one, which was already merged.

# Testing

Unit and integration (with custom tester charm).

# Changelog (aka what to ignore)

* .github/workflows/release.yaml and .github/workflows/ci.yaml - added tox targets - **safely ignore**
* .jujuignore - avoid some files in the charm
* lib/charms/data_platform_libs/v0/database_provides.py - added lib - **safely ignore**
* lib/charms/mysql/v0/mysql.py - bumped lib -  **safely ignore**
* metadata.yaml - added relation
* pyproject.toml - test markers
* src/charm.py - refactor method name and added version setting
* src/constants.py - new relation name
* src/relations/database.py - implementation from VM as is - **safely ignore**
* tests/integration/application-charm/* - tester charm from VM as is - **safely ignore**
* tests/integration/helpers.py - added helper methods for relations
* tests/integration/test_charm.py - added markers for old test
* tests/integration/test_database.py - itest from VM as is - **safely ignore**
* tests/unit/test_charm.py - fixed after addition of version testing
* tests/unit/test_database.py - utests from VM as is - **safely ignore**
* tox.ini - new tests added


tests/integration/application-charm/src/charm.py
```diff
@@ -0,0 +1,198 @@
...
+        config = {
+            "user": remote_data["username"],
+            "password": remote_data["password"],
+            "host": ro_database,
+            "database": self.database_name,
+            "raise_on_warnings": False,
+        }
+
+        with MysqlConnector(config, commit=False) as cursor:
+            rows = self._read_test_data(cursor, remote_relation.id)
```
marceloneppel, line 127: Probably it is a good idea to also test that it is really a read only endpoint by trying to write some data to it. Feel free to dismiss this idea (I know that it is being tested in some way in https://github.com/canonical/mysql-k8s-operator/blob/bc5d536d4bd3630e16386dcb440652b670637b00/tests/unit/test_database.py#L35)

marceloneppel (approved):
LGTM! I commented about an additional integration test check in the tester charm code. It seems as a better way to confirm the correct endpoints in the read only endpoints relation data field.

I also liked the idea about the logic being in the tester charm instead of in the integration tests code.

WRFitch (approved):

lib/charms/mysql/v0/mysql.py
```diff
@@ -175,6 +175,10 @@ class MySQLGetMySQLVersionError(Error):
     """Exception raised when there is an issue getting the MySQL version."""
 
 
+class MysqlGetClusterPrimaryAddressError(Error):
```
shayancanonical, line 178: minor nit: MysqlGetClusterPrimaryAddressError -> MySQLGetClusterPrimaryAddressError

shayancanonical (approved):

tests/integration/application-charm/src/charm.py
paulomach: that's very reasonable

lib/charms/mysql/v0/mysql.py
paulomach: actually, have to do before on the lib!

PR #94, opened 2022-09-27
Add mysqlrouter extra user roles for the database relation interface

shayancanonical:
# Issue
We have not yet ported over the `mysqlrouter` extra user roles in the database relation interface. This role is used to create a user used by mysqlrouter to bootstrap and additionally needs to be granted `CREATE USER` privileges.

# Solution
Port over the `mysqlrouter` extra user roles. Use `mysqlsh.cluster.setup_router_account({update: 1})` to set up the bootstrap user. Create helper method to grant `CREATE USER` privileges to this user.

# Release Notes
Add mysqlrouter extra user roles for the database relation interface

### DIFF
--- a/lib/charms/mysql/v0/mysql.py
+++ b/lib/charms/mysql/v0/mysql.py
@@ -83,7 +83,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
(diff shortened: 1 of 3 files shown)


welpaolo (approved):
Looks good

marceloneppel (approved):
Looks good!

shayancanonical:
Thanks for the reviews!

PR #97, opened 2022-10-06
Move around integration tests into appropriate files/directories

shayancanonical:
# Issue
1. We would like to move all relation tests and test application charm into its own directory (so we can cleanly add HA shakedown tests)
2. We are repeating a lot of CI workflows in `ci.yaml` and again in `release.yaml`
3. Some tests are incorrectly misplaced
4. Flake8 version is pinned (due to a prior bug), but does not need to be pinned

# Solution
1. Move all relation tests and test application charm into separate directory
2. Import CI workflows from `ci.yaml` into `release.yaml`
3. Correctly move these tests from `test_database` (relation tests) to `test_charm`
4. Unpin the flake8 version in the tox file

# Release Notes
Move around integration tests into appropriate files/directories


paulomach (approved):
thank you!

WRFitch (approved):

shayancanonical:
Thanks for the reviews!

PR #98, opened 2022-10-07
Tidy get-cluster-status action output on failure

swalladge:
`get_cluster_status()` returns None on failure,
so that must be accounted for when setting the get-cluster-status
action results dictionary.

Also ensure that this `None` return value is documented
and accounted for elsewhere in the code where it wasn't previously.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -508,7 +508,21 @@ class MySQLOperatorCharm(CharmBase):
 
     def _get_cluster_status(self, event: ActionEvent) -> None:
         """Get the cluster status without topology."""
-        event.set_results(self._mysql.get_cluster_status())
+        status = self._mysql.get_cluster_status()
+        if status:
+            event.set_results(
+                {
+                    "success": True,
+                    "status": status,
+                }
+            )
+        else:
+            event.set_results(
+                {
+                    "success": False,
+                    "message": "Failed to read cluster status.  See logs for more information.",
+                }
+            )
 
     def _restart(self, _) -> None:
         """Restart server rolling ops callback function.
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -62,7 +62,7 @@ async def get_cluster_status(ops_test: OpsTest, unit: Unit) -> Dict:
     """
     get_cluster_status_action = await unit.run_action("get-cluster-status")
     cluster_status_results = await get_cluster_status_action.wait()
-    return cluster_status_results.results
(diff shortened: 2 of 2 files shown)


src/charm.py
```diff
@@ -431,7 +431,16 @@ def _on_set_password(self, event: ActionEvent) -> None:
...
+        status = self._mysql.get_cluster_status()
+        if status:
+            event.set_results(status)
+        else:
+            event.set_results(
+                {
+                    "error": "Failed to get cluster status.  "
+                    "See logs for more information and traceback.",
+                }
+            )
```
shayancanonical, line 443: i would suggest standardizing the result structure for both success and error scenarios. 

upon success:
```
{
    "success": true,
    "status": status_dictionary
}
```

upon error:
```
{
    "success": false
}
```
this enables the caller to correctly and confidently condition on the `success` of the action

shayancanonical: I wouldn't sweat it too much if it's impossible to get the type of the field to be a boolean (at least we have a string with consistent values that we can rely on to condition upon)

shayancanonical (approved):

lib/charms/mysql/v0/mysql.py
```diff
@@ -677,14 +677,15 @@ def is_instance_in_cluster(self, unit_label: str) -> bool:
             )
             return False
 
-    def get_cluster_status(self) -> dict:
+    def get_cluster_status(self) -> Optional[dict]:
```
paulomach, line 680: @swalladge this change would be better suited on the [mysql vm operator](https://github.com/canonical/mysql-operator), that's where this library file resides.

I'm replicating it, so don't worry doing so for this change.

paulomach (approved):
lgtm - Thanks for the contribution @swalladge

lib/charms/mysql/v0/mysql.py
shayancanonical: Great catch @paulomach

paulomach:
@swalladge , refer to [this commit](https://github.com/canonical/mysql-operator/pull/56/commits/33cfeca1a8554338107d9e9c0ffcb3c1d1029b30) to fix the broken test

shayancanonical:
@swalladge The linting error can be fixed by the [following change](https://github.com/canonical/mysql-k8s-operator/pull/127/files#diff-ef2cef9f88b4fe09ca3082140e67f5ad34fb65fb6e228f119d3812261ae51449R49) (in another PR)

PR #99, opened 2022-10-07
Replicate CI job dependency + try to address flakiness of tests

shayancanonical:
# Issue
1. Alex introduced a dependency for tests (to save time and cycles running heavy integration tests if unit tests or lints fail). We do not yet have this in this repository
2. We are not raising on blocked in the `scale_application` helper function
3. Somehow, scaling the database more than one unit at a time is causing the following error trace (https://github.com/canonical/mysql-k8s-operator/actions/runs/3204633434/jobs/5236555089#step:4:541). Not suggesting that this is the root cause

# Solution
1. Replicate the dependencies for tests to this repo
2. Raise on blocked
3. Scale up one application at a time to see if that reduces the frequency of the error

# Release Notes
Replicate CI job dependency + try to address flakiness of tests

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -29,6 +29,9 @@ jobs:
 
   integration-test-microk8s-charm:
     name: Charm integration tests (microk8s)
+    needs:
+      - lint
+      - unit-test
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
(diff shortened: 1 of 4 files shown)


paulomach (approved):
looks legit

marceloneppel (approved):
What a valuable addition! I need to replicate this to the PostgreSQL charms repos.

shayancanonical:
Thanks for the amazingly quick reviews!

PR #103, opened 2022-10-12
Implement a test to validate scaling works without deletion of data

shayancanonical:
# Issue
1. The mysql lib has some updates that have not yet imported into this repository (including some bug fixes related to adding instances to clusters and escaping database identifiers)
2. We defer() the `peer_relation_joined_event` if there was an issue adding an instance to the cluster
3. We have not yet implemented a test to confirm that scaling works without deletion of data.

[jira ticket](https://warthogs.atlassian.net/browse/DPE-438)

# Solution
1. Update the mysql lib to the latest version
2. Move the unit into blocked state rather than deferring
3. Implement the test

# Release Notes
Implement the preserve data on delete replication test


WRFitch (approved):

tests/integration/high_availability/test_replication.py
```diff
@@ -255,8 +258,161 @@ async def test_no_replication_across_clusters(ops_test: OpsTest, continuous_writ
...
+        server_config_credentials["password"],
+        insert_before_scale_up_sql,
+        commit=True,
+    )
+
+    select_value_before_scale_up_sql = [
+        f"SELECT id FROM `{database_name}`.`{table_name}` WHERE id = '{value_before_scale_up}'",
+    ]
+
+    for attempt in Retrying(stop=stop_after_delay(10), wait=wait_fixed(2)):
```
paulomach, line 321: why the need for retrying in here, shouldn't the cluster be settled?

tests/integration/high_availability/test_replication.py
```diff
@@ -255,8 +258,161 @@ async def test_no_replication_across_clusters(ops_test: OpsTest, continuous_writ
...
+                    unit_address,
+                    server_config_credentials["username"],
+                    server_config_credentials["password"],
+                    select_value_before_scale_up_sql,
+                )
+                assert output[0] == value_before_scale_up
+
+    # scale up the mysql application
+    await scale_application(ops_test, mysql_application_name, 4)
+    await ensure_n_online_mysql_members(ops_test, 4)
```
paulomach, line 336: this is so handy

paulomach (approved):
Small question, otherwise lgtm

tests/integration/high_availability/test_replication.py
shayancanonical: Retrying for any replication lag that may be present

src/charm.py
```diff
@@ -360,7 +367,7 @@ def _on_peer_relation_joined(self, event: RelationJoinedEvent):
 
         except MySQLAddInstanceToClusterError:
             logger.debug(f"Unable to add instance {new_instance_fqdn} to cluster.")
-            event.defer()
```
MiaAltieri, line 363: you mentioned this was an issue in the PR description, why? Wouldn't we want to try again?

src/charm.py
```diff
@@ -349,7 +349,14 @@ def _on_peer_relation_joined(self, event: RelationJoinedEvent):
 
         # Add new instance to the cluster
         try:
-            self._mysql.add_instance_to_cluster(new_instance_fqdn, new_instance_label)
+            cluster_primary = self._mysql.get_cluster_primary_address()
+            if not cluster_primary:
+                self.unit.status = BlockedStatus("Unable to retrieve the cluster primary")
```
MiaAltieri, line 354: you mentioned this was an issue in the PR description, why? Wouldn't we want to try again and defer?

shayancanonical: here, i think it is better to return after moving the unit into blocked status because we have confirmed that the instance is configured for the innodb cluster, but we are unable to get the cluster primary's address. this usually will indicate a problem to be resolved which we wont surface if we defer

src/charm.py
shayancanonical: this i thought was a problem, because rarely does adding an instance to a cluster twice resolve the original issue (plus the instance may have some stale bad state). if we defer, we never surface the problem to the admin, which is why i think it better to move to blocked status

MiaAltieri (approved):

shayancanonical:
Thanks for the reviews!

PR #110, opened 2022-10-25
Fix/first unit restart

paulomach:
# Issue

When restarting mysqld service for a single node cluster (case when rolling restart right after cluster creation that happens in clusters), the node does not set itself as cluster primary.

# Solution

Detect deployment/state as single node cluster and `reboot from outage` when restarting.

# Context

Necessary for bundle with TLS.

# Testing

Manual validation only for now. Test to follow on bundle repo.

# Release Notes

Ignore changes to lib file, bumped only.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -513,6 +513,12 @@ class MySQLOperatorCharm(CharmBase):
         container = self.unit.get_container(CONTAINER_NAME)
         container.restart(MYSQLD_SERVICE)
 
+        # when restart done right after cluster creation (e.g bundles)
+        # or for single unit deployments, it's necessary reboot the
+        # cluster from outage to restore unit as primary
+        if self.app_peer_data["units-added-to-cluster"] == "1":
+            self._mysql.reboot_from_complete_outage()
+
         unit_label = self.unit.name.replace("/", "-")
 
         try:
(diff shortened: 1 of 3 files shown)


src/charm.py
```diff
@@ -513,6 +513,12 @@ def _restart(self, _) -> None:
         container = self.unit.get_container(CONTAINER_NAME)
         container.restart(MYSQLD_SERVICE)
 
+        # when restart done right after cluster creation (e.g bundles)
+        # or for single unit deployments, it's necessary reboot the
+        # cluster from outage to restore unit as primary
+        if self.app_peer_data["units-added-to-cluster"] == "1":
+            self._mysql.reboot_from_complete_outage()
```
MiaAltieri, line 520: might be nice to log something here

MiaAltieri (approved):
nice, looking forward to the tests 😎

shayancanonical (approved):
lgtm, but would you know why the node does not set itself as the primary? also, what about deploying from 0 to N units (where N > 1), does everything work as expected?

paulomach:
> lgtm, but would you know why the node does not set itself as the primary?

From observation, when GR enabled mysqld is restarted, the default behavior is to try to get back in the cluster as secondary. If there's other nodes, some other node will be elected as primary, and the restarted node is happily rejoined. 
When there's no other node though, the single node will not put itself as primary by itself without intervention - the reboot_from_complete_outage does the trick gracefully.

> also, what about deploying from 0 to N units (where N > 1), does everything work as expected?

Yes it does

PR #114, opened 2022-11-15
Refactor scale down logic to not blindly remove all unreachable instances from the cluster

shayancanonical:
# Issue
Since there is an issue with juju in k8s (where teardown event dont fire), we use `force remove instance` from the cluster to remove scaled down units. However, we are currently (on leader-elected and peer-relation-departed) removing all unreachable instances from the cluster. This is problematic as an instance could be unreachable due to a network partition, but should be able to rejoin the cluster once the partition has been resolved.

# Solution
Only `force remove instance`s that are scaled down. For this, we rely on `planned units` (which returns the number of planned units for an application) and the fact that juju with k8s provisions units in sequence 0, 1, 2,... Thus we will only remove any instance from the cluster whose index is >= planned units.

# Release Notes
Refactor scale down logic to not blindly remove all unreachable instances from the cluster

### DIFF
--- a/src/mysql_k8s_helpers.py
+++ b/src/mysql_k8s_helpers.py
@@ -279,7 +279,15 @@ class MySQL(MySQLBase):
             raise MySQLWaitUntilUnitRemovedFromClusterError()
 
     def force_remove_unit_from_cluster(self, unit_address: str) -> None:
-        """Force removes the provided unit from the cluster."""
+        """Force removes the provided unit from the cluster.
+
+        Args:
+            unit_address: The address of unit to force remove from cluster
+
+        Raises:
+            MySQLForceRemoveUnitFromClusterError - if there was an issue force
+                removing the unit from the cluster
+        """
         cluster_status = self.get_cluster_status()
         if not cluster_status:
             raise MySQLForceRemoveUnitFromClusterError()
--- a/src/mysql_k8s_helpers.py
+++ b/src/mysql_k8s_helpers.py
@@ -9,6 +9,7 @@ import logging
 import os
 
 from charms.mysql.v0.mysql import (
+    Error,
     MySQLBase,
     MySQLClientError,
     MySQLConfigureInstanceError,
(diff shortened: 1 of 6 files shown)


src/mysql_k8s_helpers.py
```diff
@@ -256,108 +253,75 @@ def create_custom_config_file(self, report_host: str) -> None:
...
 
     @retry(
-        retry=retry_if_result(lambda x: not x),
+        retry=retry_if_exception_type(MySQLWaitUntilUnitRemovedFromClusterError),
         stop=stop_after_attempt(10),
         wait=wait_fixed(60),
     )
-    def _wait_till_all_members_are_online(self) -> None:
-        """Wait until all members of the cluster are online.
+    def _wait_until_unit_removed_from_cluster(self, unit_address: str) -> None:
```
paulomach, line 260: Could not find this being used anywhere. I'm missing something?

tests/integration/high_availability/test_replication.py
```diff
@@ -57,13 +58,16 @@ async def test_kill_primary_check_reelection(ops_test: OpsTest, continuous_write
     client = lightkube.Client()
     client.delete(Pod, primary.name.replace("/", "-"), namespace=ops_test.model.info.name)
 
+    time.sleep(60)
```
paulomach, line 61: that's perfectly acceptable

paulomach (approved):
lgtm, left one comment that may result in code removal

src/mysql_k8s_helpers.py
```diff
@@ -256,108 +253,75 @@ def create_custom_config_file(self, report_host: str) -> None:
...
-            # In case of an error (cluster still not stable), raise an error and retry
-            logger.warning(
-                f"Failed to remove unreachable instances on {self.instance_address} with error {e.message}"
-            )
-            raise MySQLRemoveInstancesNotOnlineRetryError(e.message)
-        except RetryError as e:
+            self._run_mysqlsh_script("\n".join(remove_instance_commands))
+        except (
+            MySQLClientError,
+            MySQLWaitUntilUnitRemovedFromClusterError,
```
marceloneppel, line 319: This exception and its definition can also be removed if the method that raises it isn't being called (the one Paulo asked about).

marceloneppel (approved):
Nice approach to solve this issue, Shayan!

src/mysql_k8s_helpers.py
shayancanonical: Great catch! It was designed to be called in the function that forcefully removed a unit from the cluster. Added the function call in 0a939bd6332579e0fa0af593567f1b67265b4e8b

src/mysql_k8s_helpers.py
shayancanonical: Good catch! I just missed the call to the function that may throw this error. Added the function call in 0a939bd6332579e0fa0af593567f1b67265b4e8b

shayancanonical:
Thanks for the reviews!

PR #116, opened 2022-11-24
Implement an integration test for recovery after a network cut

shayancanonical:
# Issue
1. We are blindly force removing `unreachable` instances from the cluster. These should not be removed as they need to auto-rejoin the cluster and recover data after it is once again reachable.
2. We do not have a way to simulate network cuts. K8S network policies only filter out new connections, but do not drop packets for existing connections.
3. We do not have an integration test that tests auto recovery after a network cut.

# Solution
1. Do not force remove `unreachable` instances from the cluster. Instead, remove instances based on `planned_units` (if the `unreachable` instance index >= `planned_units`)
2. Use helm to install [chaos mesh](https://chaos-mesh.org/) to be able to simulate network cuts
3. Add integration test for auto recovery after network cuts.

# Release Notes
Implement an integration test for recovery after a network cut

### DIFF
--- a/tests/integration/high_availability/high_availability_helpers.py
+++ b/tests/integration/high_availability/high_availability_helpers.py
@@ -2,6 +2,10 @@
 # See LICENSE file for licensing details.
 
 import logging
+import os
+import string
+import subprocess
+import tempfile
 from pathlib import Path
 from typing import List, Optional, Tuple
 
@@ -19,7 +23,14 @@ from helpers import (
 )
 from juju.unit import Unit
 from pytest_operator.plugin import OpsTest
-from tenacity import RetryError, Retrying, stop_after_delay, wait_fixed
+from tenacity import (
+    RetryError,
+    Retrying,
+    retry,
+    stop_after_attempt,
+    stop_after_delay,
+    wait_fixed,
+)
 
 # Copied these values from high_availability.application_charm.src.charm
 DATABASE_NAME = "continuous_writes_database"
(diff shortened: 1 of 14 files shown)


tests/integration/high_availability/test_self_healing.py
```diff
@@ -257,3 +257,65 @@ async def test_graceful_crash_of_primary(ops_test: OpsTest, continuous_writes) -
...
+@pytest.mark.order(2)
+@pytest.mark.abort_on_fail
+@pytest.mark.self_healing_tests
+async def test_network_cut_affecting_an_instance(ops_test: OpsTest, continuous_writes) -> None:
+    """Test for a network cut affecting an instance."""
+    mysql_application_name, _ = await high_availability_test_setup(ops_test)
+
+    await ensure_all_units_continuous_writes_incrementing(ops_test)
+
+    deploy_chaos_mesh(ops_test, ops_test.model.info.name)
```
dragomirp, line 271: This could be factored out as a fixture or a setup/teardown, so that `destroy_chaos_mesh()` is not called only if the test is successful.

tests/integration/high_availability/high_availability_helpers.py
```diff
@@ -228,6 +239,45 @@ async def relate_mysql_and_application(
     )
 
 
+def deploy_chaos_mesh(ops_test: OpsTest, namespace: str) -> None:
```
dragomirp, line 242: Doesn't look like ops_test is used

tests/integration/high_availability/high_availability_helpers.py
```diff
@@ -228,6 +239,45 @@ async def relate_mysql_and_application(
...
+                namespace,
+            ]
+        ),
+        shell=True,
+        env=env,
+    )
+
+
+def destroy_chaos_mesh(ops_test: OpsTest, namespace: str) -> None:
+    """Deploy chaos mesh to the provided namespace.
```
marceloneppel, line 265:
```suggestion
    """Remove chaos mesh from the provided namespace.
```

marceloneppel (approved):
Amazing work, Shayan! Thanks for implementing this test. It'll help us a lot in the other charms.

tests/integration/high_availability/test_self_healing.py
shayancanonical: Great suggestion! Addressed in be15a67bd659c981a1705d287ef1d10a9e930f38

tests/integration/high_availability/high_availability_helpers.py
shayancanonical: Good catch! Addressed in be15a67bd659c981a1705d287ef1d10a9e930f38

tests/integration/high_availability/high_availability_helpers.py
```diff
@@ -228,6 +239,45 @@ async def relate_mysql_and_application(
...
+                "tests/integration/high_availability/scripts/deploy_chaos_mesh.sh",
+                namespace,
+            ]
+        ),
+        shell=True,
+        env=env,
+    )
+
+
+def destroy_chaos_mesh(ops_test: OpsTest, namespace: str) -> None:
```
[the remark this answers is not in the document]
shayancanonical: Addressed in be15a67bd659c981a1705d287ef1d10a9e930f38

tests/integration/high_availability/high_availability_helpers.py
shayancanonical: Good eye. Addressed in be15a67bd659c981a1705d287ef1d10a9e930f38

dragomirp (approved):

shayancanonical:
Thanks for the reviews!

paulomach:
BTW, lgtm

PR #124, opened 2022-12-06
Use `microk8s.helm3` instead of snap in CI to match local environments

carlcsaposs-canonical:


shayancanonical:
@carlcsaposs-canonical i like the change. but i did not install `helm3` using snap, but instead enabled the microk8s addon (`microk8s enable helm3`) when running the tests locally. good catch!

shayancanonical (approved):

taurus-forever (approved):
Nice!

shayancanonical (approved):
LGTM

PR #126, opened 2022-12-09
Update base from Ubuntu 20.04 to 22.04

### LINKED ISSUE #125: Release to CharmHub CI fails: no suitable 'build-on' environment
### Bug Description

From CI logs:
Run canonical/charming-actions/upload-charm@2.1.1
```shell
/usr/bin/sudo snap install charmcraft --classic --channel latest/stable
charmcraft 2.1.0 from Canonical** installed
/usr/bin/sudo charmcraft pack --destructive-mode --quiet
No suitable 'build-on' environment found in any 'bases' configuration.
```

### To Reproduce

Last passing CI run: https://github.com/canonical/mysql-k8s-operator/actions/runs/3613216881/jobs/6094949283
First failing CI run: https://github.com/canonical/mysql-k8s-operator/actions/runs/3638786440/jobs/6167166180

Note that the difference between the CI runs is:
Last passing runs on Ubuntu 20.04.5 LTS
First failing runs on Ubuntu 22.04.1 LTS

(CI job `runs-on: ubuntu-latest`)

carlcsaposs-canonical:
# Issue
Closes #125

# Release Notes
Upgrade from Ubuntu 20.04 to 22.04

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -30,7 +30,7 @@ jobs:
     needs:
       - lib-check
       - ci-tests
-    runs-on: ubuntu-22.04 # must match base in charmcraft.yaml
+    runs-on: ubuntu-latest
     steps:
       - name: Checkout
         uses: actions/checkout@v2


.github/workflows/release.yaml
```diff
@@ -30,7 +30,7 @@ jobs:
     needs:
       - lib-check
       - ci-tests
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04 # must match base in charmcraft.yaml
```
paulomach, line 33: since `latest == 22.04` what's the benefit of pinning here?
github will stick to lts only, and in worst cases, when we reach 24.04, not pinning it will be instrumental in warning about the necessity of an update.

paulomach (commented):
left a discussion point

.github/workflows/release.yaml
taurus-forever: my $0.05:

> Although if we upgrade to 24.04 before github actions we'd have to pin the version at that point

Or prepare compatibility upfront.
In general I agree with you, however mass-pinning versions without a real need have list of opposite drawbacks.
At the moment we are using 'ubuntu-latest' image everywhere, I would keep it "identical" even if not yet "perfect".

taurus-forever (approved):

MiaAltieri (approved):

PR #132, opened 2023-01-16
Remove lib-juju binding to 2.9.11 (maybe no longer necessary?)

taurus-forever:
Test ONLY.


taurus-forever:
> Test ONLY.

It is still necessary until we use juju-agent 2.9.29, we need to upgrade at least to 2.9.34.
See https://github.com/juju/python-libjuju/issues/770

PR #137, opened 2023-01-19
Update libjuju to 2.9.38.1

carlcsaposs-canonical:
# Issue

# Solution

# Context

New release versioning policy: https://github.com/juju/python-libjuju/releases/tag/2.9.38.1

# Testing

# Release Notes

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -69,7 +69,7 @@ commands =
 [testenv:integration-charm]
 description = Run charm integration tests
 deps =
-    juju==2.9.38.1 # juju 3.3.0 has issues with retrieving action results
+    juju==2.9.38.1
     lightkube
     mysql-connector-python
     pytest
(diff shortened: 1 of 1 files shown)


shayancanonical (approved):

tox.ini
```diff
@@ -69,7 +69,7 @@ commands =
 [testenv:integration-charm]
 description = Run charm integration tests
 deps =
-    juju==2.9.11 # juju 3.3.0 has issues with retrieving action results
+    juju==2.9.38.1 # juju 3.3.0 has issues with retrieving action results
```
taurus-forever, line 72: I think the comment "# juju 3.3.0 has issues with retrieving action results" makes no sense anymore and should be removed... moveover 3.3.0 is wrong and never existed. It was a comment about 3.0.0 initially.

shayancanonical: jfyi: the comment referred to the juju agent version while there is also the libjuju package which is confusingly named `juju`

taurus-forever (approved):
Yes! We are moving to new releases!!! Superb!

PR #139, opened 2023-02-02
Noop peer relation changed event handler for mysql relation if relation does not exist

shayancanonical:
# Issue
We are spitting out a lot of info statements when deferring events
See [the following pastebin](https://pastebin.canonical.com/p/P5VVP8Q4DZ/) for such logs

# Solution
Noop and return from the peer relation changed event handler if `mysql` relation does not exist. 

# Release Notes
Noop peer relation changed event handler for mysql relation if relation does not exist


paulomach (approved):
magnificent

paulomach (approved):

taurus-forever (approved):
Tnx for taking care!

PR #142, opened 2023-02-09
Fix/endpoints as k8s services

paulomach:
# Issue

The charm does not expose cluster endpoints as k8s services [DPE-1176](https://warthogs.atlassian.net/browse/DPE-1176)
`&`
The charm does not use the k8s services fqdn for the `database` relation `{read-only-}endpoints` [DPE-1175](https://warthogs.atlassian.net/browse/DPE-1175)

# Solution

Adds extra kubernetes helper module that's used by the database relation provider to manager kubernetes services and pods.

# Testing
Tested against `data-integrator` charm.
Unit tests added.
Int test with existing ones.

# Release notes (important files to review)

* `k8s_helpers.py` - new kubernetes helper class
* `mysql_provider.py` - refactor to use k8s_helpers - change method for setting and updating endpoints
* `*` - chore

[DPE-1176]: https://warthogs.atlassian.net/browse/DPE-1176?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DPE-1175]: https://warthogs.atlassian.net/browse/DPE-1175?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/k8s_helpers.py
```diff
@@ -0,0 +1,119 @@
...
+                    ports=[ServicePort(port=3306, targetPort=3306)],
+                    type="ClusterIP",
+                ),
+            )
+
+            try:
+                self.client.create(service)
+            except ApiError as e:
+                if e.status.code == 403:
+                    logger.error("Kubernetes service creation failed: `juju trust` needed")
```
taurus-forever, line 67: You need to add trust=True to tests/integration/relations/test_database.py : test_relation_broken(),
otherwise test is failing here.

taurus-forever (commented):
Asked Mykola to have a look here, as I am still learning `ops` K8s possibilities.
Test fix necessary.

src/k8s_helpers.py
```diff
@@ -0,0 +1,119 @@
...
+                if e.status.code == 409:
+                    logger.warning("Kubernetes service already exists")
+                else:
+                    logger.error("Kubernetes service creation failed: %s", e)
+                raise KubernetesClientError()
+            else:
```
shayancanonical, line 73: we can reduce the cyclomatic complexity here by moving the logger statement into the try block after `self.client.create`

src/k8s_helpers.py
```diff
@@ -0,0 +1,119 @@
...
+            service_name = f"{self.model.app.name}-{role}"
+
+            try:
+                self.client.delete(Service, service_name, namespace=self.namespace)
+            except ApiError as e:
+                if e.status.code == 403:
+                    logger.warning("Kubernetes service deletion failed: `juju trust` needed")
+                else:
+                    logger.warning("Kubernetes service deletion failed: %s", e)
+            else:
```
shayancanonical, line 92: same as above comment about cyclomatic complexity

src/k8s_helpers.py
```diff
@@ -0,0 +1,119 @@
...
+
+        try:
+            self.client.patch(Pod, pod_name or self.pod_name, pod)
+        except ApiError as e:
+            if e.status.code == 403:
+                logger.error("Kubernetes pod label creation failed: `juju trust` needed")
+            else:
+                logger.error("Kubernetes pod label creation failed: %s", e)
+            raise KubernetesClientError()
+        else:
```
shayancanonical, line 118: same comment about cyclomatic complexity

src/relations/mysql_provider.py
```diff
@@ -396,8 +303,12 @@ def _on_database_broken(self, event: RelationBrokenEvent) -> None:
             # run once by the leader
             return
 
+        if len(self.model.relations[DB_RELATION_NAME]) == 1:
+            # remove kubernetes service when last relation is removed
+            self.k8s_helpers.delete_endpoint_services(["primary", "replicas"])
```
shayancanonical, line 308: thoughts of tying the endpoint services with the charm and not the `mysql_client` relation? this way, we can [do what Mykola did](https://github.com/canonical/mysql-router-k8s-operator/blob/main/src/charm.py#L93-L131) to avoid manually deleting services

I have to test yet, but this may also have the benefit of not needing the `--trust` flag when deploying the charm

paulomach: Good call, that's a smart move.

Not sure about not having to trust though, will test it.

src/relations/mysql_provider.py
```diff
@@ -84,82 +86,32 @@ def _get_or_set_password(self, relation) -> str:
...
+            for pod in _endpoints_to_pod_list(rw_endpoints):
+                self.k8s_helpers.label_pod("primary", pod)
+            # ro pod labels
+            for pod in _endpoints_to_pod_list(ro_endpoints):
+                self.k8s_helpers.label_pod("replicas", pod)
+        except MySQLGetClusterEndpointsError as e:
+            logger.exception("Failed to get cluster members", exc_info=e)
+        except KubernetesClientError:
+            logger.debug("Can't update pod labels")
+            # TODO: Block?
```
delgod, line 114: yes, block

delgod (approved):

src/k8s_helpers.py
paulomach: addressed #b3c73bf

src/k8s_helpers.py
paulomach: addressed #b3c73bf

src/k8s_helpers.py
```diff
@@ -0,0 +1,119 @@
...
+            try:
+                self.client.create(service)
+            except ApiError as e:
+                if e.status.code == 403:
+                    logger.error("Kubernetes service creation failed: `juju trust` needed")
+                if e.status.code == 409:
+                    logger.warning("Kubernetes service already exists")
+                else:
+                    logger.error("Kubernetes service creation failed: %s", e)
+                raise KubernetesClientError()
```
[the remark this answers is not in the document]
paulomach: It's to keep the pattern that we've been using of raising custom error and avoid one more import on `mysql_provider`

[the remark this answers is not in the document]
paulomach: technically yes, but I don't like it as a pattern, since obfuscate the real origin

PR #145, opened 2023-02-14
Opportunistic test improvements

paulomach:
## Issue

1. Often on CI, there's failures due mysqld restarting with `super-read-only` flag enabled *before* being added to the cluster, on initialization. The culprit seems to be incomplete shutdowns, as pebble allows only 5s before killing a service.
2. Continuous writes should be stupid and never stop trying to write to DB.

## Solution

1. Send a `SIGTERM` to workload before a `container.restart`
2. Make continuous write test app dumbly write and not hang when lost connection


src/mysql_k8s_helpers.py
```diff
@@ -173,6 +173,10 @@ def configure_instance(self) -> None:
         try:
             super(MySQL, self).configure_instance(restart=False)
 
+            # ensure clean shutdown
+            # TODO: remove when pebble allow more time for shutdown
+            #       https://github.com/canonical/pebble/pull/190
+            self.container.exec(["pkill", "-15", "mysqld"])
```
shayancanonical, line 179: actually, this may become unnecessary when we use `mysqld_safe` instead of `mysqld` (when we address https://warthogs.atlassian.net/browse/DPE-1293)

shayancanonical (approved):
This is wonderful!

tests/integration/high_availability/application_charm/src/connector.py
```diff
@@ -2,13 +2,19 @@
 # Copyright 2022 Canonical Ltd.
 # See LICENSE file for licensing details.
 
+import signal
+
 import mysql.connector
 
 
+def timeout_handler(signum, frame):
+    raise Exception("Query timed out")
```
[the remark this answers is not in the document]
paulomach: tbh, neither of those. It should use a suitable builtin or a custom one. I'll change to the builtin:
```suggestion
    raise TimeoutError
```

src/mysql_k8s_helpers.py
taurus-forever: I do not think mysqld_safe will help... as pebble will kill mysqld_safe->mysqld after 5s anyway.

tests/integration/high_availability/application_charm/src/continuous_writes.py
```diff
@@ -37,13 +37,14 @@ def continuous_writes(database_config: Dict, table_name: str, starting_number: i
...
                 )
         except mysql.connector.errors.DatabaseError as e:
-            # errno 2003: can't connect to mysql database
-            if e.errno != 2003:
-                raise
+            if e.errno == 1062:
+                with MySQLConnector(database_config) as cursor:
+                    cursor.execute(f"SELECT max(number) FROM `{table_name}`")
+                    result = cursor.fetchall()
+                    next_value_to_insert = result[0][0] + 1
```
taurus-forever, line 44: LGTM, could be simplified the next time by using column with `AUTO_INCREMENT` and simply insert anything and checking primary key value (always in innodb_buffer_pool).

src/mysql_k8s_helpers.py
taurus-forever: AFAIK, pkill exits immediately after sending the signal to the mysqld and the next line in code is calling pebble to stop/kill mysqld: `self.container.restart("mysqld")`

Q: how do we increase/avoid 5s pebble timeout here?

taurus-forever (approved):
IMHO, It is generally wrong...  but LGTM to keep moving with tests stabilization (hope to replace it with pebble timeout soon).

@delgod can we rely on some `mysqladmin ping/status/extended-status` to check db unit/cluster is fully (re)started (instead of sending SIGTERM)? Tnx!

src/mysql_k8s_helpers.py
shayancanonical: what i meant is that we will no longer need to run `self.container.restart("mysqld")` when we switch to using mysqld_safe. the `mysqlsh.configure_instance` function will invoke `RESTART` itself when it is done configuring the instance

PR #167, opened 2023-03-02
Remove `_set_group_replication_initial_variables`

carlcsaposs-canonical:
## Issue
Super read only error when local address was set after unit is configured for group replication and before it joins the cluster

[DPE-1365](https://warthogs.atlassian.net/browse/DPE-1365)

## Solution
This configuration appears to be related to the `XCOM` communication stack. We use the `MYSQL` communication stack, so this configuration shouldn't be necessary.

## Context
https://dev.mysql.com/doc/mysql-shell/8.0/en/configuring-cluster-instance-ports.html

[DPE-1365]: https://warthogs.atlassian.net/browse/DPE-1365?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/mysql_k8s_helpers.py
+++ b/src/mysql_k8s_helpers.py
@@ -215,10 +215,6 @@ class MySQL(MySQLBase):
             self.container.restart(MYSQLD_SERVICE)
             logger.debug("Waiting until MySQL to restart")
             self.wait_until_mysql_connection()
-
-            # set global variables to enable group replication in k8s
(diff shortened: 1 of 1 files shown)


shayancanonical (approved):

paulomach (approved):
that feels good!

PR #172, opened 2023-03-15
Update AWS bucket for backups integration tests

shayancanonical:
## Issue
- We have updated the credentials for AWS to use the shared team account bucket (`data-charms-testing`). However, we have not yet updated our tests to use this bucket.
- The data_interface charm library is out of date

## Solution
- Update backups integration tests to use `data-charms-testing` bucket and `mysql-k8s` as the path to store the backups
- Update the data_interface charm library in the mysql-k8s charm and the test application charms as well


paulomach (approved):

taurus-forever (approved):
Tnx!

PR #176, opened 2023-03-17
Fix new units addition after failure.

paulomach:
## Issue

Transient issue where adding new units to a cluster with a to-be recovered primary has not yet recovered.
As described [DPE-1224](https://warthogs.atlassian.net/browse/DPE-1224).

## Solution

Sets maintenance status for all cases (unit configured or not), which will defer new units joining the cluster until `Active Status` is reached.
Speed up recovery when restarting from a pre-configured attached storage.

### Noteworthy

During testing, sometimes the databag went missing and I could see [this bug in action](https://bugs.launchpad.net/juju/+bug/2011277)

[DPE-1224]: https://warthogs.atlassian.net/browse/DPE-1224?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -358,6 +358,7 @@ class MySQLOperatorCharm(CharmBase):
         if not self._prepare_configs(container):
             return
 
+        self.unit.status = MaintenanceStatus("Initialising mysqld")
         if self.unit_peer_data.get("unit-configured"):
             # Only update pebble layer if unit is already configured for GR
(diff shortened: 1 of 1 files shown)


taurus-forever (approved):
Nice!

PR #179, opened 2023-03-22
Replace mysqld with mysqld_safe

shayancanonical:
[dpe-1293](https://warthogs.atlassian.net/browse/DPE-1293)

## Issue
We run `mysqld` as a pebble layer, and `mysqld` need to be able to correctly restart itself. It cannot when it is run with pebble, since `mysqld` restarts a systemd service to restart itself. We are currently manually restarting `mysqld` with `pebble restart`

## Solution
Replace the usage of `mysqld` with `mysqld_safe`, and get rid of the restart with pebble.


src/mysql_k8s_helpers.py
```diff
@@ -785,17 +744,17 @@ def safe_stop_mysqld(self):
 
         def get_mysqld_pid(self):
             try:
-                process = self.container.exec(["pgrep", "-x", "mysqld"])
+                process = self.container.exec(["pgrep", "-x", MYSQLD_SAFE_SERVICE])
```
taurus-forever, line 747: This will return the pid of mysqld_safe and not mysqld, while function is get_mysqld_pid().
Is it fine? I believe on `RESTART` mysqld PID will be changed but mysqld_safe PID will remain.

paulomach: Additionally, does `mysqld_safe` fork `mysqld` process (being it parent process) or exec it (replacing the context)?

paulomach (approved):
LGTM, after we clear @taurus-forever  point on `RESTART` under `mysqld_safe`

PR #181, opened 2023-03-27
WIP: Update the upstream-source to the charmed-mysql rock

shayancanonical:
## Issue
Our charm still relies on an image built with the Dockerfile, but we would like to switch to the ROCK based off the charmed-mysql snap

## Solution
Update the upstream source for the OCI in metadata.yaml


shayancanonical:
Closing in favor of a new PR (without the change to install dependencies upon instnace configuration)

PR #197, opened 2023-04-06
Documentation Sync with Discourse

deusebio:


docs/how-to/h-enable-encryption.md
```diff
@@ -0,0 +1,53 @@
...
+
+* generate external keys for each unit
+
+```
+openssl genrsa -out external-key-0.pem 3072
+openssl genrsa -out external-key-1.pem 3072
+openssl genrsa -out external-key-2.pem 3072
+```
+
+* apply both private keys on each unit, shared internal key will be allied only on juju leader
```
paulomach, line 34:
```suggestion
* apply both private keys on each unit, shared internal key will be applied only on juju leader
```

docs/how-to/h-enable-encryption.md
```diff
@@ -0,0 +1,53 @@
+# How to enable encryption
+
```
paulomach, line 2: Can we add something like this?

> MySQL will enable encrypted connections by default with self generated certificates. Though also by default, connecting clients can disable encryption by setting the connection ssl-mode as disabled.
> When related with the `tls-certificates-operator` the charmed operator for MySQL will require that every client connection (new and running connections) use encryption, rendering an error when attempting to establish an unencrypted connection.

docs/how-to/h-manage-app.md
```diff
@@ -0,0 +1,54 @@
...
+
+Also extended permissions can be requested using `mysql-root` edpoint:
+```shell
+juju relate mysql-k8s:mysql-root wordpress-k8s
+```
+
+
+## Rotate applications password
+
+To rotate the passwords of users created for related applications, the relation should be removed and related again. That process will generate a new user and password for the application.
```
paulomach, line 33:
```suggestion
To rotate the passwords of users created for related applications, the relation should be removed and related again. That process will generate a new user and password for the application, while retaining the requested database and data.
```

docs/how-to/h-manage-units.md
```diff
@@ -0,0 +1,35 @@
+# How to deploy and manage units
+
+## Basic Usage
+
+To deploy a single unit of MySQL using its default configuration
+```shell
+juju deploy mysql-k8s --channel edge
```
paulomach, line 7:
```suggestion
juju deploy mysql-k8s --channel edge --trust
```

docs/how-to/h-manage-units.md
```diff
@@ -0,0 +1,35 @@
...
+```
+
+It is customary to use MySQL with replication. Hence usually more than one unit (preferably an odd number to prohibit a "split-brain" scenario) is deployed. To deploy MySQL with multiple replicas, specify the number of desired units with the `-n` option.
+```shell
+juju deploy mysql-k8s --channel edge -n <number_of_replicas>
```
paulomach, line 12:
```suggestion
juju deploy mysql-k8s --channel edge --trust -n <number_of_replicas>
```

paulomach (changes requested):
Left some suggestions/fixes here and there

taurus-forever (approved):

paulomach (approved):

PR #200, opened 2023-04-11
Pin the charmed-mysql rock to a hash

shayancanonical:
## Issue
We are using a ROCK label `charmed-mysql:8.0.32-22.04_edge` in the charm. However, a breaking a change in a new version of the ROCK can result in a broken charm. Additionally, it is possible for 2 nodes in the same cluster to have different versions of the ROCK (if a new version of the ROCK is released before a new node is added to the cluster).

## Solution
Pin the ROCK to a specific hash in the metadata file. We can add a renovate configuration to update this hash in a future PR (to automatically update and create a PR when there is a new version of the ROCK).

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -26,7 +26,7 @@ resources:
   mysql-image:
     type: oci-image
     description: Ubuntu LTS Docker image for MySQL
-    upstream-source: ghcr.io/canonical/charmed-mysql:8.0.32-22.04_edge
+    upstream-source: ghcr.io/canonical/charmed-mysql:8.0.32-22.04_edge@sha256:924cb913733002e0936a36b6edab0bb8a4d72f1b92be0e0b7a933e797b7ef215
 peers:
   database-peers:
     interface: mysql_peers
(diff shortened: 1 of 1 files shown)


shayancanonical:
> > Additionally, it is possible for 2 nodes in the same cluster to have different versions of the ROCK
> 
> Is this possible? I thought Charmhub charm releases had a pinned OCI image revision

I thought that it was possible if we release a new OCI image to charmhub and the new pod does not have the image locally (and so needs to pull the OCI image from charmhub). Unless juju has a way of pinning a charm revision with the resource revision

paulomach:
> > Unless juju has a way of pinning a charm revision with the resource revision
> 
> I believe that when you release a charm to a channel, you specify the resource revision (alongside the charm revision): https://github.com/canonical/data-platform-workflows/blob/0c8efe4f4266dd2934daff32222e2132e50f47b0/release_charm.py#L98
> 
> > I feel like it may still be a good idea to pin the image in the code
> 
> Agree

That's the case for released charms. Pinning the digest will be useful for better tracking deployments and for CI/CD testing, since the later uses the resources as declared in the `metadata.yaml`

taurus-forever (approved):
We might return to human readable versioning by reventing this.
Hovewer, meanwhile, let's stay on the safe side and pin ROCK OCI.

PR #201, opened 2023-04-11
Mark `test_no_replication_across_clusters` as unstable

carlcsaposs-canonical:
## Issue
Test failed 4 times in a row, blocking release CI: https://github.com/canonical/mysql-k8s-operator/actions/runs/4667631230/jobs/8265947160

## Solution
Mark test as unstable

### DIFF
--- a/tests/integration/high_availability/test_replication.py
+++ b/tests/integration/high_availability/test_replication.py
@@ -53,6 +53,7 @@ async def test_check_consistency(ops_test: OpsTest, continuous_writes) -> None:
 
 
 @pytest.mark.abort_on_fail
+@pytest.mark.unstable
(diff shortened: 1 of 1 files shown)


paulomach (changes requested):
@carlcsaposs-canonical the pr #203 maybe can help on this, since what seems to be going on is OOM scenario.
If so, we should change the test to include memory constrains to the deployment, e.g.:
The equivalent of:
```
juju deploy --constraints mem=1G <charm_file>
```
into `ops_test.deploy`

taurus-forever (changes requested):
Agree to Paulo [here](https://github.com/canonical/mysql-k8s-operator/pull/201#pullrequestreview-1381581748) , we should constraint RAM usage (in all our tests) and not marking test unstable.

Issue #205, opened 2023-04-13
Stabilize CI tests on GH by applying RAM constraints on controller/model level (to limit each MySQL unit to 1GB max)


taurus-forever:
Hi,

GH tests crashes with various reasoning often caused by lack of RAM and a lot of units running on free GitHub runners.
MySQL charm code allocates `innodb_buffer_pool_size` to 50% of container/vm RAM (4GB in fact).
Each container's RAM is 100% of GH Runner RAM (8GB each).
Each VM's RAM is also 100% of host RAM (by default).

In fact on GH Runner with 8GB RAM, we are starting two MySQL clusters, with three units each, setting innodb_buffer_pool_size=4GB for each unit... 

We must limit RAM to 1GB maximum for each unit. It should be more then enough for all our tests.

See details [here](https://github.com/canonical/mysql-k8s-operator/issues/202#issuecomment-1505126043) and [here](https://github.com/canonical/operator/issues/844#issuecomment-1488268566).

P.S. maybe we should request `charmed-kubernetes/actions-operator` to set some constraints by default as most of tests do not need a lot of RAM, but looking for tests stability.

taurus-forever:
Unfortunately, it is no possible to limit Juju app RAM usago on controller bootstrap:
https://bugs.launchpad.net/juju/+bug/2016242

@paulomach please consider to lock contraints on juju test model creation:
> subprocess.check_output("juju set-model-constraints cores=2 mem=1G")

PR #203, opened 2023-04-12
add support for memory constrained deployment

paulomach:
## Issue

The charm does not respect resources limits.
 **Extra context**: the method to calculate memory limits is very grabby and uses host system memory as base.
Further tuning may be necessary on k8s as the k8s cluster node workload tends to be more diverse, requiring a somewhat less grabby calculation.

## Solution

Refactor method that retrieve total memory to check for resources limits.
Chore:
- moved KubernetesHelpers reference from relation provider to charm in order to refer the object broadly
- refactor/added unit tests

Refer to [DPE-1598](https://warthogs.atlassian.net/browse/DPE-1598)
Closes #192 

[DPE-1598]: https://warthogs.atlassian.net/browse/DPE-1598?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #206, opened 2023-04-13
DPE-1656 Limit tests RAM usage by every unit/container to 1GB

taurus-forever:
## Issue
Often and strange tests crashes on GH runner while all is OK on localhost.

## Solution
GH Runners have 8GB RAM only. We are launching 1-2 MySQL clusters with three units each and allocating 50% for innodb_buffer_pull_size. OOM is an often trouble in this case.

Close: https://github.com/canonical/mysql-k8s-operator/issues/205

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -75,7 +75,7 @@ jobs:
         with:
           provider: microk8s
           # This is needed until https://bugs.launchpad.net/juju/+bug/1977582 is fixed
-          bootstrap-options: "--agent-version 2.9.29"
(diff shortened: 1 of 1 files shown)


paulomach (approved):
lgtm - I have added to the PR #203   [this line](https://github.com/canonical/mysql-k8s-operator/blob/13d1f2af0a6e8f4f4a59dacf5b2b6df1668c4e7b/tests/integration/high_availability/high_availability_helpers.py#L153) that effectively achieve the same (a bit more limited memory), but just for mysql charms.
Should this PR approach super seed that?

PR #210, opened 2023-04-17
Use data dir for temp backup dir + format list-backups output as table

shayancanonical:
[dpe-1597](https://warthogs.atlassian.net/browse/DPE-1597)
counterpart of [vm pr 185](https://github.com/canonical/mysql-operator/pull/185)

## Issue
1. We are not correctly formatting the output of the list-backups action
2. We are currently only displaying the successfully "finished" backup ids in the list-backups output
3. We are not using # to being temp backup/restore directories while performing backups/restores. Additionally, the temp backup directory is not in the MySQL data directory (thus we could run out of disk space)

## Solution
1 and 2. Update the `s3_helpers`, `mysql` and `backups` charm libraries to pull the in the latest code from the vm charm. These will address the output format of the list-backups action.
3. Use # for the temp restore directories. Additionally, use the data directory as the temp backup directory.


taurus-forever (approved):
Tnx!

PR #215, opened 2023-04-21
Remove upgrade_user_for_mysql_router

carlcsaposs-canonical:
The mysql router charm will setup its own router user

### DIFF
--- a/lib/charms/mysql/v0/mysql.py
+++ b/lib/charms/mysql/v0/mysql.py
@@ -122,6 +122,10 @@ class MySQLCheckUserExistenceError(Error):
     """Exception raised when checking for the existence of a MySQL user."""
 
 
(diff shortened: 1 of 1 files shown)


lib/charms/mysql/v0/mysql.py
```diff
@@ -415,54 +407,6 @@ def does_mysql_user_exist(self, username: str, hostname: str) -> bool:
             )
             raise MySQLCheckUserExistenceError(e.message)
 
-    def configure_mysqlrouter_user(
```
shayancanonical, line 418: i think we need this helper method for the `db-router` relation in the VM charm

shayancanonical (approved):

taurus-forever (approved):
LGTM, cross-check with @delgod

delgod (approved):

PR #224, opened 2023-05-10
Cleanup router metadata

carlcsaposs-canonical:
Depends on canonical/mysql-k8s-operator#220 (change base branch after canonical/mysql-k8s-operator#220 is merged)

## Issue
Router metadata not removed when router leaves relation

## Solution
Remove metadata when router leaves relation

### DIFF
--- a/lib/charms/mysql/v0/mysql.py
+++ b/lib/charms/mysql/v0/mysql.py
@@ -632,7 +632,7 @@ class MySQLBase(ABC):
             raise MySQLDeleteUserError("Unable to query cluster primary address")
         drop_user_command = [
             f"shell.connect('{self.server_config_user}:{self.server_config_password}@{primary_address}')",
-            f'session.run_sql("DROP USER `{username}`")',
+            f"session.run_sql(\"DROP USER `{username}`@'%'\")",
         ]
         try:
             self._run_mysqlsh_script("\n".join(drop_user_command))


src/relations/mysql_provider.py
```diff
@@ -54,6 +56,10 @@ def __init__(self, charm) -> None:
         self.framework.observe(
             self.charm.on[DB_RELATION_NAME].relation_broken, self._on_database_broken
         )
+        self.framework.observe(
+            self.charm.on[DB_RELATION_NAME].relation_departed,
+            self._on_database_provides_relation_departed,
+        )
```
delgod, line 62: maybe some comments good to add for better readability
```suggestion
        # cleanup users created by router for router<->app relation
        self.framework.observe(
            self.charm.on[DB_RELATION_NAME].relation_broken, self._on_database_broken
        )
        
        # cleanup router metadata and the router user
        self.framework.observe(
            self.charm.on[DB_RELATION_NAME].relation_departed,
            self._on_database_provides_relation_departed,
        )
```

src/relations/mysql_provider.py
```diff
@@ -309,7 +315,38 @@ def _on_database_broken(self, event: RelationBrokenEvent) -> None:
...
+                f"More than one router user for departing unit {event.departing_unit.name}"
+            )
+            return
+
+        user = users[0]
+        try:
+            self.charm._mysql.delete_user(user.username)
+            logger.info(f"Deleted router user {user.username}")
+        except MySQLDeleteUserError:
+            logger.error(f"Failed to delete user {user.username}")
```
delgod, line 347: maybe we should defer/retry if the deletion fails.
up to @paulomach

src/relations/mysql_provider.py
```diff
@@ -309,7 +317,38 @@ def _on_database_broken(self, event: RelationBrokenEvent) -> None:
         relation_id = event.relation.id
         try:
             self.charm._mysql.delete_users_for_relation(relation_id)
-            logger.info(f"Removed user for relation {relation_id}")
+            logger.info(f"Removed user(s) for relation {relation_id}")
         except MySQLDeleteUsersForRelationError:
-            logger.error(f"Failed to delete user for relation {relation_id}")
+            logger.error(f"Failed to delete user(s) for relation {relation_id}")
+            return
```
shayancanonical, line 323: is this return necessary? isnt going to return since it's the end of the method?

src/relations/mysql_provider.py
```diff
@@ -309,7 +317,38 @@ def _on_database_broken(self, event: RelationBrokenEvent) -> None:
...
+        users = self.charm._mysql.get_mysql_router_users_for_unit(
+            relation_id=event.relation.id, mysql_router_unit_name=event.departing_unit.name
+        )
+        if not users:
+            return
+
+        if len(users) > 1:
+            logger.error(
+                f"More than one router user for departing unit {event.departing_unit.name}"
+            )
```
shayancanonical, line 341: isnt it possible to have more than 1 user returned (if there are multiple applications related to the mysqlrouter application)?

lib/charms/mysql/v0/mysql.py
```diff
@@ -528,6 +549,32 @@ def _get_statements_to_delete_users_with_attribute(
...
         ]
 
+    def get_mysql_router_users_for_unit(
+        self, *, relation_id: int, mysql_router_unit_name: str
+    ) -> list[RouterUser]:
+        """Get users for related MySQL Router unit.
+
+        For each user, get username & router ID attribute.
+        """
+        primary_address = self.get_cluster_primary_address()
```
paulomach, line 559: Do we really need to run this query from the primary?

lib/charms/mysql/v0/mysql.py
```diff
@@ -578,9 +625,40 @@ def delete_users_for_relation(self, relation_id: int) -> None:
...
             raise MySQLDeleteUsersForRelationError(e.message)
 
+    def delete_user(self, username: str) -> None:
+        """Delete user."""
+        primary_address = self.get_cluster_primary_address()
+        if not primary_address:
+            raise MySQLDeleteUserError("Unable to query cluster primary address")
+        drop_user_command = [
+            f"shell.connect('{self.server_config_user}:{self.server_config_password}@{primary_address}')",
+            f'session.run_sql("DROP USER `{username}`")',
```
paulomach, line 638: Any reason to not drop users with a `DROP USER IF EXISTS '{username}'@'%';` ?

src/relations/mysql_provider.py
paulomach: I'd agree - if there's a error dropping a user it maybe the case to retry - I've also commented on the lib method to make it harder to fail.

paulomach: Sure

paulomach (approved):
lgtm

delgod (approved):

PR #232, opened 2023-05-24
DPE-1892 DPE-1702 Use special user for backups + miscellaneous backups improvements

shayancanonical:
## Issue
We have made some backup related improvements in the vm charm, but have yet to port them over to the k8s charm. Improvements include:
1. Using a special backups user
2. Logging backups commands executed with debug log
3. Setting a flag to catch all pipe failures when executing commands
4. Input sanitation for values incoming from s3 integrator or from backups related actions

## Solution
Port changes to the k8s charm


paulomach (approved):
lgtm - Question: Does xtrabackup creates any log file? If so, we should make it write to a know location and add it to COS integration

shayancanonical:
xtrabackup does not create a log file. we log the command output in debug log with level `DEBUG`, and we also upload these logs to S3

shayancanonical:
Going to merge with knowledge of failing self-healing tests (as we are hoping [PR 225](https://github.com/canonical/mysql-k8s-operator/pull/225/files) helps with the failures)

PR #243, opened 2023-06-17
DPE-2124 More memory saves

paulomach:
Same as vm [PR229](https://github.com/canonical/mysql-operator/pull/229)


taurus-forever (commented):
Didn't we consider to go with "profile=devel"?

paulomach:
> Didn't we consider to go with "profile=devel"?

es! this was before the decision was made, I'll just close it.

PR #274, opened 2023-08-09
DPE-2352 Start mysqld-exporter on COS relation and restart upon monitoring password change

shayancanonical:
## Issue
1. We are always running the mysqld-exporter, even if the relation with COS does not exist
2. We are not restarting mysqld-exporter when the password for the monitoring user changes

## Solution
1. Run `pwd` (or no-op) command in the mysql-exporter pebble service when no relation to COS exists. When a relation is formed, run the mysqld-exporter in the same pebble service (using pebble service override strategy)
2. Replan the pebble services upon monitoring user password changes (to run mysqld-exporter with the most up to date password)

### DIFF
--- a/lib/charms/mysql/v0/mysql.py
+++ b/lib/charms/mysql/v0/mysql.py
@@ -88,6 +88,7 @@ from constants import (
     BACKUPS_USERNAME,
     CLUSTER_ADMIN_PASSWORD_KEY,
     CLUSTER_ADMIN_USERNAME,
+    COS_AGENT_RELATION_NAME,
     MONITORING_PASSWORD_KEY,
     MONITORING_USERNAME,
     PASSWORD_LENGTH,
@@ -110,7 +111,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 38
+LIBPATCH = 39
 
 UNIT_TEARDOWN_LOCKNAME = "unit-teardown"
 UNIT_ADD_LOCKNAME = "unit-add"
@@ -403,6 +404,12 @@ class MySQLCharmBase(CharmBase):
         self._mysql.update_user_password(username, new_password)
 
         self.set_secret("app", secret_key, new_password)
+
+        if (
+            username == MONITORING_USERNAME
+            and len(self.model.relations.get(COS_AGENT_RELATION_NAME, [])) > 0
+        ):
+            self._mysql.restart_mysql_exporter()
 
     def _get_cluster_status(self, event: ActionEvent) -> None:
         """Action used  to retrieve the cluster status."""
@@ -2293,6 +2300,11 @@ Swap:     1027600384  1027600384           0
     @abstractmethod
     def start_mysqld(self) -> None:
         """Starts the mysqld process."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def restart_mysql_exporter(self) -> None:
+        """Restart the mysqld exporter."""
         raise NotImplementedError
 
     @abstractmethod
--- a/src/charm.py
+++ b/src/charm.py
@@ -47,6 +47,7 @@ from constants import (
     CLUSTER_ADMIN_USERNAME,
     CONTAINER_NAME,
     GR_MAX_MEMBERS,
+    COS_AGENT_RELATION_NAME,
     MONITORING_PASSWORD_KEY,
     MONITORING_USERNAME,
     MYSQL_LOG_FILES,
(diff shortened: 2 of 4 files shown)


src/charm.py
```diff
@@ -134,43 +147,50 @@ def _mysql(self) -> MySQL:
...
+                "command": "/start-mysqld-exporter.sh",
+                "startup": "enabled",
+                "user": MYSQL_SYSTEM_USER,
+                "group": MYSQL_SYSTEM_GROUP,
+                "environment": {
+                    "DATA_SOURCE_NAME": (
+                        f"{MONITORING_USERNAME}:"
+                        f"{self.get_secret('app', MONITORING_PASSWORD_KEY)}"
+                        f"@unix({MYSQLD_SOCK_FILE})/"
+                    ),
```
paulomach, line 189: why not just make `startup` a function of cos_relation, i.e.:
```suggestion
                MYSQLD_EXPORTER_SERVICE: {
                    layer["services"][MYSQLD_EXPORTER_SERVICE] = {
                        "override": "replace",
                        "summary": "mysqld exporter",
                        "command": "/start-mysqld-exporter.sh",
                        "startup":  "enabled" if self._has_cos_relation else "disabled",
                        "user": MYSQL_SYSTEM_USER,
                        "group": MYSQL_SYSTEM_GROUP,
                        "environment": {
                            "DATA_SOURCE_NAME": (
                                f"{MONITORING_USERNAME}:"
                                f"{self.get_secret('app', MONITORING_PASSWORD_KEY)}"
                                f"@unix({MYSQLD_SOCK_FILE})/"
                            ),
```

shayancanonical: addressed in deecb90ea97e5fdeb7c0ec31d93943338540d81a. `startup` only controls if the service should be started when pebble starts. we also need to `container.stop()` to inactivate the service

shayancanonical: not entirely sure, but i saw replan() called but the `mysqld-exporter` service still enabled. https://ops.readthedocs.io/en/latest/#ops.Container.replan indicates possibly that it only restarts running services, and starts enabled services.

shayancanonical: that is indeed correct: `replan()` did not stop the service:

```
root@mysql-0:/# pebble plan
services:
    mysqld_exporter:
        summary: mysqld exporter
        startup: disabled
        override: replace
        command: /start-mysqld-exporter.sh
        environment:
            DATA_SOURCE_NAME: monitoring:jj7ojopaCYnnyHZgcNcF8sNc@unix(/var/run/mysqld/mysqld.sock)/
        user: mysql
        group: mysql
    mysqld_safe:
        summary: mysqld safe
        startup: enabled
        override: replace
        command: mysqld_safe
        user: mysql
        group: mysql
        kill-delay: 24h0m0s
    promtail:
        summary: promtail
        startup: disabled
        override: replace
        command: /opt/promtail/promtail-static-amd64 -config.file=/etc/promtail/promtail_config.yaml
root@mysql-0:/# pebble services
Service          Startup   Current  Since
mysqld_exporter  disabled  active   today at 18:48 UTC
mysqld_safe      enabled   active   today at 18:47 UTC
promtail         disabled  active   today at 18:48 UTC
```

paulomach (approved):
nice

paulomach (dismissed):

paulomach (approved):

lib/charms/mysql/v0/mysql.py
```diff
@@ -975,6 +1020,38 @@ def remove_router_from_cluster_metadata(self, router_id: str) -> None:
...
+
+        Raises:
+            MySQLSetVariableError
+        """
+        if not instance_address:
+            instance_address = self.instance_address
+        logger.debug(f"Setting {variable} to {value} on {instance_address}")
+        set_var_command = [
+            f"shell.connect('{self.server_config_user}:{self.server_config_password}@{instance_address}')",
+            f"session.run_sql(\"SET {'PERSIST' if persist else 'GLOBAL'} {variable}={value}\")",
```
taurus-forever, line 1046: Side comment: we must be careful here, to prevent SQL injection:
> value="something; drop database mysql;"

As I can see the method used in MySQL VM tests only... but who knows in the future.

lib/charms/mysql/v0/mysql.py
```diff
@@ -1579,6 +1659,37 @@ def get_cluster_primary_address(
...
+
+        Raises:
+            MySQLSetClusterPrimaryError: If the cluster primary could not be set
+        """
+        logger.debug(f"Setting cluster primary to {new_primary_address}")
+
+        set_cluster_primary_commands = (
+            f"shell.connect_to_primary('{self.server_config_user}:{self.server_config_password}@{self.instance_address}')",
+            f"cluster = dba.get_cluster('{self.cluster_name}')",
+            f"cluster.set_primary_instance('{new_primary_address}')",
```
taurus-forever, line 1685: Q: @paulomach why don't we use a single mysql-shell call?
```
f"dba.get_cluster('{self.cluster_name}').set_primary_instance('{new_primary_address}')"
```

IMHO Pros: 2x faster (no unnecessary network roundtrips) and less code.
Cons: long Python line (and less readability) :-)

taurus-forever (dismissed):

lib/charms/mysql/v0/mysql.py
paulomach: speed wise, i don't think it changes, since you're still doing two API calls.
i do like chaining methods though, even from readability point of view.
I'm taking a note to implement on the next lib change

lib/charms/mysql/v0/mysql.py
paulomach: https://warthogs.atlassian.net/browse/DPE-2399

lib/charms/mysql/v0/mysql.py
paulomach: https://warthogs.atlassian.net/browse/DPE-2399

taurus-forever (approved):

PR #288, opened 2023-08-17
DPE-2408 Use fallback key for passwords in legacy relations

shayancanonical:
## Issue
We cannnot use `_` in secret keys (for password keys in legacy relations). Some applications already deployed will use `_` in peer relation secrets, but when secrets are rotated, they must be rotated with `-` in juju secrets rather than `_`

## Solution
Use `fallback_key` with `_` (as done in mysql vm [PR 243](https://github.com/canonical/mysql-operator/pull/243/files#diff-a7f1a6521d2009b594920ca3bb3ecec524f52fced6c7670d281ed5b335a608f6R504-R524)), but first try to store/get secrets with `-`

### DIFF
--- a/src/relations/mysql.py
+++ b/src/relations/mysql.py
@@ -20,7 +20,13 @@ from ops.charm import (
 from ops.framework import Object
 from ops.model import ActiveStatus, BlockedStatus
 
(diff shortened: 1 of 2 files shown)


paulomach (dismissed):

paulomach (approved):

taurus-forever (approved):
Tnx!

PR #294, opened 2023-08-23
Use 60s interval for fast forward instead of 5s

carlcsaposs-canonical:
Ported from https://github.com/canonical/mysql-operator/pull/306


shayancanonical (dismissed):

paulomach (dismissed):

shayancanonical (approved):

paulomach (approved):

taurus-forever (approved):
I would go with a constant to control it from one place, but up-2-you here.

PR #295, opened 2023-08-25
DPE-2178 Update mysql charm lib to v0.44 - to stop configuring mysql user root@%

shayancanonical:
## Issue
We updated the `mysql` charm lib with some feature, one of which includes not configuring the `root@%` user. We have not yet imported this charm lib into this charm.

## Solution
Import the most up to data version of the `mysql` charm lib


paulomach (approved):

taurus-forever (approved):
Thank you!

PR #302, opened 2023-09-05
[bugfix] defer reconciling pebble layer for exporter

paulomach:
## Issue

COS relation creation/departure is reconciling the pebble layer, which will hinder the charm initialization in case it happens before other initial events (leader-election, pebble-ready).
This scenario will happens on relating with COS right from the deployment, e.g. in a bundle.

## Solution

Defer handler to run after initialization.


phvalguima (approved):
That fix resolved the race condition between COS and mysql cluster

LGTM

PR #308, opened 2023-09-13
DPE-2397 allow deployment on older then 2.9.44 juju controllers

paulomach:
## Issue

pebble bundled with controllers older then 2.9.44 does not support `kill-delay` 

## Solution

Enable `kill-delay` only if current controller supports it 

_NOTE_: juju does check the assumes on the refresh - so this wont upgrade on a <2.9.44 controller.

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -75,7 +75,7 @@ assumes:
   - k8s-api
   - any-of:
       - all-of:
-          - juju >= 2.9.44
+          - juju >= 2.9.29
           - juju < 3
       - all-of:
           - juju >= 3.1.5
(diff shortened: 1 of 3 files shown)


shayancanonical (approved):

taurus-forever (dismissed):

paulomach:
> > are we supporting controllers < 2.9.44?
> > https://github.com/canonical/mysql-k8s-operator/blob/d73010c9eba7641a80b6b91982d468eaaba91a47/metadata.yaml#L78
> > 
> > TBF, this may be completely uncessary
> 
> We never tried on Juju <2.9.29. It doesn't harm here but declares supported limits. @paulomach please update it here https://charmhub.io/mysql-k8s/docs/r-requirements after merging!

Yes. IIRC, we must settle to something newer due teardown issues. I'll get back to this later

PR #311, opened 2023-09-25
Remove method duplication

paulomach:
## Issue

`configure_user` exists in shared library and has no difference.
(It used to have when we were using `root@%` to connect to database)

## Solution

Remove it


shayancanonical (approved):
:rocket:

paulomach:
This should be completed in the scope of task dpe-2178

PR #315, opened 2023-10-05
[DPE-2699] mysql-test-app safe on dp libs v18

juditnovak:
## Issue

`mysql-test-app`, used as a pre-built charm in relation tests has to be also updated to use Relational Secrets -- in order to have Juju3 tests using actual Relational Secrets instead of falling back to the databag.

## Solution

Verification of the action being safe, then https://github.com/canonical/mysql-test-app/pull/14


juditnovak:
POC demonstrated purpose, https://github.com/canonical/mysql-test-app/pull/14 merged, closing

PR #330, opened 2023-10-24
[DPE-2837] Additional testing for external secrets

juditnovak:
## Issue

Currently this behavior is not tested.

https://github.com/canonical/mysql-operator/blob/main/src/relations/mysql_provider.py#L190

(I know it coz I broke it ;-)  With green pipelines... :-( )

## Solution

1. In case we consider this behavior not needed: REMOVE the code
2. In case we consider the behavior as imporatnt: gotta be tested.

### DIFF
--- a/tests/integration/relations/test_database.py
+++ b/tests/integration/relations/test_database.py
@@ -10,7 +10,13 @@ import pytest
 import yaml
 from pytest_operator.plugin import OpsTest
 
-from ..helpers import get_relation_data, is_relation_broken, is_relation_joined
+from ..helpers import (
+    get_relation_data,
+    get_secret_data,
+    is_relation_broken,
+    is_relation_joined,
+    scale_application,
+)
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


tests/integration/relations/test_database.py
```diff
@@ -122,6 +128,50 @@ async def test_relation_creation(ops_test: OpsTest):
...
+async def test_password_preserved_in_relation(ops_test: OpsTest):
+    """Check if an empty relation is caching password."""
+    relation_data = await get_relation_data(ops_test, APPLICATION_APP_NAME, "database")
+    old_password = relation_data[0]["application-data"]["password"]
+
+    async with ops_test.fast_forward("60s"):
+        await scale_application(ops_test, DATABASE_APP_NAME, 0)
+
+    async with ops_test.fast_forward("60s"):
+        await scale_application(ops_test, DATABASE_APP_NAME, 2)
```
juditnovak, line 143: Await missing!!!!!!!!!

juditnovak:
To be addressed in the bug that it's "duplicating", see https://warthogs.atlassian.net/browse/DPE-2837

PR #333, opened 2023-11-02
[DPE-2885] Secret labels

juditnovak:
Secrets using labels

### DIFF
--- a/lib/charms/mysql/v0/mysql.py
+++ b/lib/charms/mysql/v0/mysql.py
@@ -78,6 +78,13 @@ from abc import ABC, abstractmethod
 from typing import Any, Dict, Iterable, List, Optional, Tuple, Union
 
 import ops
+from charms.data_platform_libs.v0.data_secrets import (
+    APP_SCOPE,
+    UNIT_SCOPE,
+    Scopes,
+    SecretCache,
+    generate_secret_label,
+)
 from ops.charm import ActionEvent, CharmBase, RelationBrokenEvent
 from ops.model import Unit
 from tenacity import (
@@ -116,7 +123,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 51
+LIBPATCH = 53
 
 UNIT_TEARDOWN_LOCKNAME = "unit-teardown"
 UNIT_ADD_LOCKNAME = "unit-add"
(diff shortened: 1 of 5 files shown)


lib/charms/mysql/v0/mysql_secrets.py
```diff
@@ -0,0 +1,142 @@
...
+# Increment this PATCH version before using `charmcraft publish-lib` or reset
+# to 0 if you are raising the major API version
+LIBPATCH = 1
+
+APP_SCOPE = "app"
+UNIT_SCOPE = "unit"
+Scopes = Literal[APP_SCOPE, UNIT_SCOPE]
+
+
+class PostgresSQLSecretsError(Exception):
```
shayancanonical, line 26: should be MySQLSecretsError

juditnovak: :facepalm: 

Sure I'll fix that :-)

lib/charms/mysql/v0/mysql.py
```diff
@@ -535,36 +540,43 @@ def has_cos_relation(self) -> bool:
...
-            content = secret.get_content()
-            self.app_secrets = content
-            logger.debug(f"Retrieved secret {key} for app from juju")
+        if not secret:
+            logger.debug("Getting a secret when secret is not added in juju")
+            return
 
-        return self.app_secrets.get(key)
+        value = secret.get_content().get(key)
+        logger.debug(f"Retrieved secret {key} for unit from juju")
```
shayancanonical, line 578: can we remove this debug statement? as i understand it, it will log every time `_get_secret_from_juju` is called, which is going to be a lot

tests/integration/relations/test_database.py
```diff
@@ -122,6 +128,50 @@ async def test_relation_creation(ops_test: OpsTest):
...
+
+    async with ops_test.fast_forward("60s"):
+        await scale_application(ops_test, DATABASE_APP_NAME, 2)
+
+    relation_data = await get_relation_data(ops_test, APPLICATION_APP_NAME, "database")
+    secret_uri = relation_data[0]["application-data"]["secret-user"]
+    secret_data = await get_secret_data(ops_test, secret_uri)
+    new_password = secret_data["password"]
+
+    assert new_password == old_password
```
shayancanonical, line 172: should we have these integration tests (since we're scaling to 0 and then back to 2 on a relation)? i know in VM, we do not reuse volumes, but i dont know if the PVC gets reused in K8s

shayancanonical (dismissed):

shayancanonical (approved):

delgod (approved):

shayancanonical (changes requested):
Actually, we should make this change in the VM charm first (as the `lib/charms/mysql/v0/mysql.py` lib is updated and published there first). Going to replicate the changes to [the following PR](https://github.com/canonical/mysql-operator/pull/348)

shayancanonical (approved):

delgod (approved):

PR #360, opened 2024-01-19
Update MySQL to 8.0.35

carlcsaposs-canonical:
https://warthogs.atlassian.net/browse/DPE-3227


delgod:
> @delgod What is your vision here?

All the tests passed, so we don't need to change anything right now.
Please let me know if I am wrong.
Later, we can improve/review security and make some system users "localhost" only if possible.

paulomach (approved):

Issue #310, opened 2023-09-25
Charm stuck in `Initialising mysqld`


gboutry:
## Steps to reproduce

1. Deploy sunbeam

## Expected behavior
Charm should become ready

## Actual behavior

Charm has an unhandled exception and never becomes ready

## Versions

Operating system: Ubuntu 22.04.3 LTS

Juju CLI: 3.2.3-genericlinux-amd64

Juju agent: 3.2.3

Charm revision: 99 (8.0/candidate)

microk8s: MicroK8s v1.26.8 revision 5883

## Log output
```
2023-09-25T11:48:38.238Z [container-agent] 2023-09-25 11:48:38 INFO juju.worker.uniter.operation runhook.go:186 ran "start" hook (via hook dispatching script: dispatch)
2023-09-25T11:48:39.462Z [container-agent] 2023-09-25 11:48:39 INFO juju.worker.uniter.operation runhook.go:186 ran "database-relation-changed" hook (via hook dispatching script: dispatch)
2023-09-25T11:48:40.972Z [container-agent] 2023-09-25 11:48:40 INFO juju.worker.uniter.operation runhook.go:186 ran "database-relation-joined" hook (via hook dispatching script: dispatch)
2023-09-25T11:48:42.028Z [container-agent] 2023-09-25 11:48:42 INFO juju.worker.uniter.operation runhook.go:186 ran "database-relation-changed" hook (via hook dispatching script: dispatch)
2023-09-25T11:48:43.152Z [container-agent] 2023-09-25 11:48:43 INFO juju.worker.uniter.operation runhook.go:186 ran "database-relation-changed" hook (via hook dispatching script: dispatch)
2023-09-25T11:48:44.168Z [container-agent] 2023-09-25 11:48:44 INFO juju.worker.uniter.operation runhook.go:186 ran "database-relation-changed" hook (via hook dispatching script: dispatch)
2023-09-25T11:48:45.239Z [container-agent] 2023-09-25 11:48:45 INFO juju.worker.uniter.operation runhook.go:186 ran "database-relation-joined" hook (via hook dispatching script: dispatch)
2023-09-25T11:48:46.475Z [container-agent] 2023-09-25 11:48:46 INFO juju.worker.uniter.operation runhook.go:186 ran "upgrade-relation-changed" hook (via hook dispatching script: dispatch)
[...]

PR #361, opened 2024-01-19
Re-throw the exception when checking mysqlsh connection

motjuste:
## Issue

When connection to `mysqlsh` is checked, any `MySQLClientError` is _swallowed_.  This does not help with debuggability.

## Solution

Change the API from returning a `bool` to returning nothing, but promising the faithfully throw an exception while checking `mysqlsh` connection if one occurred, _without_ losing any information that can help with debugging.  The call-site can then do the same, better here with a higher-level error, but again, without losing any details of the original error.

### DIFF
--- a/lib/charms/mysql/v0/mysql.py
+++ b/lib/charms/mysql/v0/mysql.py
@@ -2550,18 +2550,13 @@ class MySQLBase(ABC):
             logger.exception("Failed to kill external sessions")
             raise MySQLKillSessionError
 
-    def check_mysqlsh_connection(self) -> bool:
+    def check_mysqlsh_connection(self) -> None:
         """Checks if it is possible to connect to the server with mysqlsh."""
         connect_commands = (
             f"shell.connect('{self.server_config_user}:{self.server_config_password}@{self.instance_address}')",
             'session.run_sql("SELECT 1")',
         )
-
-        try:
-            self._run_mysqlsh_script("\n".join(connect_commands))
-            return True
-        except MySQLClientError:
-            return False
+        self._run_mysqlsh_script("\n".join(connect_commands))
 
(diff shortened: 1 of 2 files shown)


motjuste:
I am not sure why some of the Juju 2.9 integration tests fail as I do not believe I have modified any code related to those failures.  The integration tests for Juju 3.1.6 pass.

motjuste:
The problem I am trying to debug is the issue #310 on Juju 3.1.6 on top of microk8s 1.24 that I have been banging my head against while trying to deploy the MLFlow bundle.  I hope that this change will help me debug that issue better.

paulomach (commented):
@motjuste , this is reasonable. Before going fwd take a look at my comment in the [issue](https://github.com/canonical/mysql-k8s-operator/issues/310#issuecomment-1906122179).

Also, we would need to update the `lib/charms/mysql/v0/mysql.py` `LIBPATCH` field and sync this change with the `mysql-operator` (VM) charm

paulomach:
@motjuste , latest merge [added exception logging](https://github.com/canonical/mysql-k8s-operator/blob/4cd27ff365c3ff4c937274fbdf4b2db02d8dabce/lib/charms/mysql/v0/mysql.py#L2495) - if you believe it's enough I would like to close this PR

motjuste:
> @motjuste , latest merge [added exception logging](https://github.com/canonical/mysql-k8s-operator/blob/4cd27ff365c3ff4c937274fbdf4b2db02d8dabce/lib/charms/mysql/v0/mysql.py#L2495) - if you believe it's enough I would like to close this PR

Hi ... sorry for late response ... looks good, since `logger.exception` will also add the exception info.

PR #367, opened 2024-01-26
DPE-2919 update pod labels on preemptive switchover

### LINKED ISSUE #335: Preemptive primary switchover does not update pod labels
When preemptively changing primary, the charm does not update Pod labels on the same handler, which will disrupt client connections for more time that's actually necessary, since the k8s services used to map the primary and replicas will be outdated until some other hook updates it.

Todo:

* on preemptive primary switchover, call k8s `update_pod_labels` method
* [optional] test k8s service endpoint for primary before exiting the hook.

paulomach:
## Issue

* current preemptive switchover does not update pod labels (for service selection)

## Solution

* update pod labels on successful switchover

Closes #335

### DIFF
--- a/src/k8s_helpers.py
+++ b/src/k8s_helpers.py
@@ -113,6 +113,8 @@ class KubernetesHelpers:
             pod_name: (optional) name of the pod to label, defaults to the current pod
         """
         try:
+            if not pod_name:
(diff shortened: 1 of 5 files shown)


taurus-forever (approved):
Hey! Nice!

PR #385, opened 2024-03-01
DPE-3389 support rollback incompat data dir 

paulomach:
## Issue

Upgrade rollback fails when new mysqld modify data_dir without backward compatibility.

## Solution

Add treatment for the case that data_dir is modified, for multi-unit cluster (reset and clone)

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -498,7 +498,7 @@ class MySQLOperatorCharm(MySQLCharmBase, TypedCharmBase[CharmConfig]):
 
         # Add the pebble layer
         logger.debug("Adding pebble layer")
-        container.add_layer(MYSQLD_SAFE_SERVICE, self._pebble_layer, combine=False)
+        container.add_layer(MYSQLD_SAFE_SERVICE, self._pebble_layer, combine=True)
         container.restart(MYSQLD_SAFE_SERVICE)
 
         logger.debug("Waiting for instance to be ready")
@@ -553,7 +553,8 @@ class MySQLOperatorCharm(MySQLCharmBase, TypedCharmBase[CharmConfig]):
             return
 
         if not self.upgrade.idle:
-            # pebble ready task delegated to upgrade
+            # when upgrading pebble ready is
+            # task delegated to upgrade code
             return
 
         container = event.workload
--- a/src/charm.py
+++ b/src/charm.py
@@ -552,6 +552,10 @@ class MySQLOperatorCharm(MySQLCharmBase, TypedCharmBase[CharmConfig]):
             event.defer()
             return
 
+        if not self.upgrade.idle:
+            # pebble ready task delegated to upgrade
+            return
+
         container = event.workload
         self._write_mysqld_configuration()
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -18,8 +18,6 @@ from charms.mysql.v0.mysql import (
     BYTES_1MB,
     MySQLAddInstanceToClusterError,
     MySQLCharmBase,
-    MySQLConfigureInstanceError,
-    MySQLConfigureMySQLUsersError,
     MySQLCreateClusterError,
     MySQLGetClusterPrimaryAddressError,
     MySQLGetMemberStateError,
@@ -72,11 +70,7 @@ from constants import (
 )
 from k8s_helpers import KubernetesHelpers
 from log_rotate_manager import LogRotateManager
-from mysql_k8s_helpers import (
-    MySQL,
-    MySQLCreateCustomConfigFileError,
-    MySQLInitialiseMySQLDError,
-)
+from mysql_k8s_helpers import MySQL
 from relations.mysql import MySQLRelation
 from relations.mysql_provider import MySQLProvider
 from relations.mysql_root import MySQLRootRelation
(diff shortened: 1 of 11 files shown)


src/charm.py
```diff
@@ -485,45 +479,45 @@ def _open_ports(self) -> None:
...
-                    container.start(MYSQLD_EXPORTER_SERVICE)
-        except (
-            MySQLConfigureInstanceError,
-            MySQLConfigureMySQLUsersError,
-            MySQLInitialiseMySQLDError,
-            MySQLCreateCustomConfigFileError,
-        ) as e:
-            logger.debug("Unable to configure instance: {}".format(e))
-            return False
+    def _configure_instance(self, container) -> None:
```
paulomach, line 491: changed to just raise exception on error, instead of boolean return

src/mysql_k8s_helpers.py
```diff
@@ -229,10 +226,13 @@ def wait_until_mysql_connection(self, check_port: bool = True) -> None:
         Retry every 5 seconds for 30 seconds if there is an issue obtaining a connection.
         """
         if not self.container.exists(MYSQLD_SOCK_FILE):
-            raise MySQLServiceNotRunningError()
+            raise MySQLServiceNotRunningError
 
-        if check_port and not self.check_mysqlsh_connection():
-            raise MySQLServiceNotRunningError("Connection with mysqlsh not possible")
+        try:
+            if check_port and not self.check_mysqlsh_connection():
```
paulomach, line 232: fix to catch exception in `self.check_mysqlsh_connection`  call

taurus-forever (approved):
Wow, this looks promising! Eager to try it in edge!

src/mysql_k8s_helpers.py
```diff
@@ -852,3 +852,7 @@ def set_cluster_primary(self, new_primary_address: str) -> None:
         """Set the cluster primary and update pod labels."""
         super().set_cluster_primary(new_primary_address)
         self.update_endpoints()
+
+    def fetch_error_log(self) -> Optional[str]:
+        """Fetch the MySQL error log."""
+        return self.read_file_content("/var/log/mysql/error.log")
```
shayancanonical, line 858: can this be a bad idea? if the error log is large, it will be loaded from disk into memory

instead, may be consider iterating through lines of the file and searching for the error code you are looking for

src/upgrade.py
```diff
@@ -301,3 +329,27 @@ def _check_server_upgradeability(self) -> None:
...
+        """Check error log for unsupported downgrade.
+
+        https://dev.mysql.com/doc/mysql-errors/8.0/en/server-error-reference.html
+        """
+        if log_content := self.charm._mysql.fetch_error_log():
+            return "MY-013171" in log_content
+
+        return False
+
+    def _reset_on_unsupported_downgrade(self, container: Container) -> None:
```
shayancanonical, line 343: thoughts on adding a simple check that this is not done if there is only 1 unit in the cluster?

tests/integration/high_availability/test_upgrade.py
```diff
@@ -124,7 +120,16 @@ async def test_upgrade_from_edge(ops_test: OpsTest, continuous_writes) -> None:
     assert leader_unit is not None, "No leader unit found"
 
     logger.info("Resume upgrade")
-    await juju_.run_action(leader_unit, "resume-upgrade")
+    while get_sts_partition(ops_test, MYSQL_APP_NAME) == 2:
+        # resume action sometime fails in CI, no clear reason
```
shayancanonical, line 124: not happy to have this workaround

tests/integration/high_availability/test_upgrade.py
```diff
@@ -190,7 +195,17 @@ async def test_fail_and_rollback(ops_test, continuous_writes, built_charm) -> No
     )
 
     logger.info("Resume upgrade")
-    await juju_.run_action(leader_unit, "resume-upgrade")
+    while get_sts_partition(ops_test, MYSQL_APP_NAME) == 2:
+        # resume action sometime fails in CI, no clear reason
```
shayancanonical, line 199: same here, not happy about this workaround

src/mysql_k8s_helpers.py
paulomach: I considered that, but given that we are rotating it (and inhibiting the rotation solely to the duration of the upgrade), it should not be enormous (as in MB)

src/upgrade.py
paulomach: that's done on the calling method (L256)

shayancanonical (approved):

tests/integration/high_availability/test_upgrade.py
paulomach: Me neither, but I should be clearer here, since I know the reason.
the run action recently changed behavior, raising when the action fails.
The action will fail when we the leader unit is the one target as the next one to be upgrade, e.g. when we have three units (0, 1*, 2) when resuming upgrade after unit 2 is completed, `resume-upgrade` will run on unit 1(leader) and set the partition to `1`, immediately rolling the pod, before the charm is able to respond the action hook

PR #419, opened 2024-05-24
Adds new option to allow configuration of binlog expire

sombrafam:


paulomach:
@sombrafam as discussed offline, we understand the need for a sane default here. Discussion is open internally and we will come back to this when time allow us. For now, I'm closing it.

PR #430, opened 2024-06-11
pin pytest-asyncio

paulomach:
**Chore**

* updated osm-bundle due issue with old bundle and newer juju
* bump libs

### DIFF
--- a/lib/charms/data_platform_libs/v0/upgrade.py
+++ b/lib/charms/data_platform_libs/v0/upgrade.py
@@ -285,7 +285,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 16
+LIBPATCH = 17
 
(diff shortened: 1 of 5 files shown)


shayancanonical (approved):
just out of curiosity and history: why this change?

paulomach:
> just out of curiosity and history: why this change?
@shayancanonical 

It's related to the pytest-asyncio update that I'd tried to fix on [dpw](https://github.com/canonical/data-platform-workflows/pull/186) 
But, in details:
* systems with python 3.12 (noble) pulled pytest-asyncio-0.27.3 on poetry update
* pytest-asyncio-0.27.3 has the bug that duplicates all pytest markers
  * pytest-asyncio add a asyncio to all `async def test...` when configure to `auto` mode, which saves the hassle of adding the mark to every test function manually
* duplicated group marker is breaking an assertion on test collection step on dpw groups plugins (hence the temptive fix on it)
* when testing the fix on dpw, tests are stuck on initial tests. Local tests showed that downgrading only pytest-asyncio fix the issue, so there's more issues in it then the markers duplication

taurus-forever (approved):

Issue #407, opened 2024-04-18
Evicted secondary node fail to cluster again


gboutry:
## Steps to reproduce

- repro.sh: https://pastebin.ubuntu.com/p/rvNYg4cHtf/
- remove_node.py: https://pastebin.ubuntu.com/p/QmjPfJqxFx/

## Expected behavior
When the pod is scheduled to another node, it should be able to join back the quorum.

## Actual behavior

The MySQL unit fails to join back the quorum, complaining about `An instance with label 'squirrel-mysql-0' is already part of this InnoDB cluster`

## Versions

Operating system: Ubuntu 22.04.4 LTS

Juju CLI: 3.4.2-genericlinux-amd64

Juju agent: 3.4.2

Charm revision: 127

microk8s: MicroK8s v1.28.7 revision 6532

## Log output
```log
2024-04-18T08:55:29.833Z [container-agent] verbose: 2024-04-18T08:55:29Z: Loading startup files...                                                                                                                                                       
2024-04-18T08:55:29.833Z [container-agent] verbose: 2024-04-18T08:55:29Z: Loading plugins...                                                                                                                                                             
2024-04-18T08:55:29.833Z [container-agent] verbose: 2024-04-18T08:55:29Z: Connecting to MySQL at: clusteradmin@squirrel-mysql-1.squirrel-mysql-endpoints.rodents.svc.cluster.local                                                                       
2024-04-18T08:55:29.833Z [container-agent] verbose: 2024-04-18T08:55:29Z: Shell.connect: tid=257: CONNECTED: squirrel-mysql-1.squirrel-mysql-endpoints.rodents.svc.cluster.local                                                                         
2024-04-18T08:55:29.833Z [container-agent] verbose: 2024-04-18T08:55:29Z: Connecting to MySQL at: mysql://clusteradmin@squirrel-mysql-1.squirrel-mysql-endpoints.rodents.svc.cluster.local:3306?connect-timeout=5000                                     
2024-04-18T08:55:29.833Z [container-agent] verbose: 2024-04-18T08:55:29Z: Dba.get_cluster: tid=258: CONNECTED: squirrel-mysql-1.squirrel-mysql-endpoints.rodents.svc.cluster.local:3306                                                                  
[...]

PR #433, opened 2024-06-13
[DPE-4118] Address drained units rejoining the cluster with a new PV

shayancanonical:
## Issue
https://github.com/canonical/mysql-k8s-operator/issues/407
When a node is drained along with the pods on that node + the PV of the unit is deleted, then when the unit is restarted, the unit tries to rejoin the cluster. However, there is already metadata for that unit in the cluster metadata that prevents the unit's SST.

## Solution
If the instance is already in non-online state when trying to join the cluster, run `cluster.rescan()` on the primary node so that the rejoining unit's SST completes without errors.


taurus-forever (approved):
The charm logic is up to Paulo here, the code is LGTM. Tnx!

src/charm.py
```diff
@@ -361,6 +361,9 @@ def join_unit_to_cluster(self) -> None:
                 # Stop GR for cases where the instance was previously part of the cluster
                 # harmless otherwise
                 self._mysql.stop_group_replication()
+                self._mysql.rescan_cluster_if_instance_already_in_cluster(
```
paulomach, line 364: In addition to [comment on the lib](https://github.com/canonical/mysql-operator/pull/462#discussion_r1638086525) :

```suggestion
                if self._mysql.is_instance_in_cluster(...
                  # ensure joining instance is not present in stale cluster metadata
                  self._mysql.rescan(..
```

paulomach (changes requested):
Complementary comment.
Also, since we could reproduce by deleting the PV, let's include a integration test for it.

src/charm.py
shayancanonical: addressed in [ecd1150](https://github.com/canonical/mysql-k8s-operator/pull/433/commits/ecd1150121fdf62bcb2f87f86d5eda7a43e775b7)

shayancanonical:
Added an integration test in [ecd1150](https://github.com/canonical/mysql-k8s-operator/pull/433/commits/ecd1150121fdf62bcb2f87f86d5eda7a43e775b7)

paulomach (approved):

src/charm.py
```diff
@@ -359,6 +364,16 @@ def join_unit_to_cluster(self) -> None:
                 # Stop GR for cases where the instance was previously part of the cluster
                 # harmless otherwise
                 self._mysql.stop_group_replication()
+
+                # If instance already in cluster, rescan cluster on primary before
+                # attempting to add the instance to the cluster
+                cluster_status = self._mysql.get_cluster_status(from_instance=cluster_primary)
+                instances = cluster_status["defaultreplicaset"]["topology"].keys()
+                if instance_label in instances:
+                    self._mysql.rescan_cluster(
```
paulomach, line 373: I wonder if the rescan is doing some harm here, since can remove not only this instance but others. Maybe use `remove_instance` api method instead?

taurus-forever (approved):
Nice fix!

src/charm.py
shayancanonical: addressed in [113db7d](https://github.com/canonical/mysql-k8s-operator/pull/433/commits/113db7d74b195d2a4b03c69ec185b6c3614430ae)

PR #444, opened 2024-06-25
DPE-4699 backup logging and other fixes

paulomach:
VM [pr#471](https://github.com/canonical/mysql-operator/pull/471) + [pr#478](https://github.com/canonical/mysql-operator/pull/478)

plus:
* fixes missing `@property` from `s3_integrator_relation_exists`, migrated to lib
* backup test splited in 3 (aws, gcp, ceph)


lib/charms/mysql/v0/mysql.py
```diff
@@ -3097,6 +3121,29 @@ def kill_unencrypted_sessions(self) -> None:
...
+                "SELECT processlist_id FROM performance_schema.threads WHERE "
+                "type = 'FOREGROUND' and connection_type is not NULL and processlist_id != CONNECTION_ID();"
+                '")'
+            ),
+            "process_id_list = [id[0] for id in processes.fetch_all()]",
+            'for process_id in process_id_list:\n  session.run_sql(f"KILL CONNECTION {process_id}")',
+        )
+
+        try:
+            self._run_mysqlsh_script("\n".join(kill_connections_command))
```
taurus-forever, line 3142: I suspect: all clients will just reconnect :-) (expected mysql client library behaviour).
What you are searching is probably MySQL Offline mode:
https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_offline_mode
https://dev.mysql.com/blog-archive/automatic-member-fencing-with-offline_mode-in-group-replication/

taurus-forever (approved):
1) see me comment about killing clients.
2) I expect backup tests separation without the 300+ lines copy&paste 3 times. As the difference there in confutation only.
3) I am in love in UX imrovements. THANK YOU! This is impressive!

lib/charms/mysql/v0/mysql.py
paulomach: [feb3b46](https://github.com/canonical/mysql-k8s-operator/pull/444/commits/feb3b46c692dbae5635f12c9e33e10e10b0fa740)

shayancanonical (approved):
I agree with Alex, I think we should look for a way to not repeat test code 3 times for each S3 storage type

PR #465, opened 2024-07-22
DPE-3557 Workaround pvc transient error

### LINKED ISSUE #378: `create Pod mysql-k8s-1 in StatefulSet mysql-k8s failed error: failed to create PVC mysql-k8s-database-3ed48296-mysql-...` error in integration test
Found while looking at Allure reports: https://docs.google.com/spreadsheets/d/1Mc8PwHlgib0Ys535MoOVTVGazD0DDVAZ46ajJIDqljs/edit#gid=306045415&range=A2

Doesn't show up as exception in juju debug-log, potentially an issue with Juju/microk8s

Workflow run: https://github.com/canonical/mysql-k8s-operator/actions/runs/7689835900/job/20952950353

Logs (for archival purposes)
[logs-intergration-test-charm-microk8s-juju-3.1.7-high_availability-test_self_healing.py-group-1.zip](https://github.com/canonical/mysql-k8s-operator/files/14240189/logs-intergration-test-charm-microk8s-juju-3.1.7-high_availability-test_self_healing.py-group-1.zip)

juju status pytest-operator output different from log file (probably changed in time between statuses checked)
screenshot

Allure report: https://canonical.github.io/mysql-k8s-operator/1554/#suites/4c5d09a1a64e6a96d4d66ccc157cefbd/2e1cec9f38179955/

paulomach:
## Issue

Integration tests often fails on a [PVC failing to be created](https://github.com/canonical/mysql-k8s-operator/actions/runs/9983902668/job/27592545064#step:22:149)

## Solution

Workaround it by not raise error on wait_for_idle when run after mysql-k8s deploy/scale commands (i.e. when PVC is created)

Credits to @lucasgameiroborges 

Fixes #378

### DIFF
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -265,6 +265,7 @@ async def scale_application(
                 timeout=(15 * 60),
                 wait_for_exact_units=desired_count,
                 raise_on_blocked=True,
+                raise_on_error=False,
             )
 
 
--- a/tests/integration/high_availability/high_availability_helpers.py
+++ b/tests/integration/high_availability/high_availability_helpers.py
@@ -165,6 +165,7 @@ async def deploy_and_scale_mysql(
             status="active",
             raise_on_blocked=True,
             timeout=TIMEOUT,
+            raise_on_error=False,
         )
 
         assert len(ops_test.model.applications[mysql_application_name].units) == num_units
--- a/tests/integration/high_availability/test_async_replication.py
+++ b/tests/integration/high_availability/test_async_replication.py
@@ -109,11 +109,13 @@ async def test_build_and_deploy(
             apps=[MYSQL_APP1],
             status="active",
             timeout=10 * MINUTE,
+            raise_on_error=False,
         ),
         second_model.wait_for_idle(
             apps=[MYSQL_APP2],
             status="active",
             timeout=10 * MINUTE,
+            raise_on_error=False,
         ),
     )
 
--- a/tests/integration/high_availability/test_upgrade.py
+++ b/tests/integration/high_availability/test_upgrade.py
@@ -60,6 +60,7 @@ async def test_deploy_latest(ops_test: OpsTest) -> None:
         apps=[MYSQL_APP_NAME, TEST_APP_NAME],
         status="active",
         timeout=TIMEOUT,
+        raise_on_error=False,
     )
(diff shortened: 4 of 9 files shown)


shayancanonical (commented):
I am a little worried about the charms going into error state (for reasons other than PVC creation), and then eventually update_status coming along and setting the status as `active` and us not catching this case due to the `raise_on_error=False`

taurus-forever:
If this really helps, why error state still failing [this](https://github.com/canonical/mysql-k8s-operator/actions/runs/10048233474/job/27772422035#step:24:123). 

Was it permanent error state and not temporary one?

taurus-forever (approved):

lucasgameiroborges:
Hi guys, here to provide some context for why this change was adopted over on PostgreSQL side:

> If this really helps, why error state still failing [this](https://github.com/canonical/mysql-k8s-operator/actions/runs/10048233474/job/27772422035#step:24:123).

@taurus-forever It seems the change was missing in this particular `wait_for_idle`: https://github.com/canonical/mysql-k8s-operator/blob/ea997403bb0689886e7f5ac55888535d8304eb6c/tests/integration/test_charm.py#L65C1-L71C10

> I am a little worried about the charms going into error state (for reasons other than PVC creation), and then eventually update_status coming along and setting the status becoming active from retries and us not catching this case due to the raise_on_error=False

@shayancanonical Me and Dragomir (who originally proposed this change in the first place on my PR, I had a different approach at first) discussed the implications of not raising errors on the `wait_for_idle` calls, the main point being: It is not that we'll completely ignore the errors, it is just that they won't get raised in the very first `wait_for_idle` after charm deploy. Sure, transient errors like the PVC one might get ignored completely, but persistent errors will either get the `wait_for_idle` to timeout due to charm never getting `active` status, or something breaking further ahead in the test. About `the status becoming active from retries`, wouldn't this mean the charm actually recovered during said retries? Unless the hooks are setting `active` status when they shouldn't.

In any case, this is not a proper fix by any means, but more of a workaround to avoid annoying transient errors right after charm deploy on CI.

paulomach:
> I am a little worried about the charms going into error state (for reasons other than PVC creation), and then eventually ~update_status coming along and setting~ the status becoming `active` from retries and us not catching this case due to the `raise_on_error=False`

As Lucas said, we will still will catch the errors, but when timeout is reached.
IMO if we can reach a better stability, It's a valid workaround.

PR #481, opened 2024-08-13
DPE-4933 Avoid early calls to unit_initialized

paulomach:
## Issue

Recent changes from the way we check for `unit_initilized` is causing pebble errors when pebble is not yet ready on the workload container.
More clearly, `unit_initialized` was moved from a databag flag to a query to the mysqld itself, meaning if that unit was added to the current cluster.

## Solution

1. Ensure calls to `unit_initialized` are only made if the workload container is ready (`can_connect() == True`).
2. During unrelated tests, It was noticed that default mysqld restart timeout value (30s) isn't enough for real world cases, so it's increased to 3600s.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -445,7 +445,8 @@ class MySQLOperatorCharm(MySQLCharmBase, TypedCharmBase[CharmConfig]):
         self.unit.status = MaintenanceStatus("restarting MySQL")
         container = self.unit.get_container(CONTAINER_NAME)
         if container.can_connect():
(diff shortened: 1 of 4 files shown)


shayancanonical (approved):

taurus-forever (approved):
The 3600s could bite us back with new findings... anyway let's try.
Tnx for a quick fix here!

paulomach:
> The 3600s could bite us back with new findings... anyway let's try. Tnx for a quick fix here!

 Make it bigger then? 30s is definitively not enough

PR #483, opened 2024-08-16
Fix metrics-endpoint created on scale up

paulomach:
## Issue

While testing, found a bug where scaling up after COS relationship is established.
The handler for `metrics-endpoint-relation-created` is running before the new unit is initialized (i.e. before pebble ready). Since the hook does `_reconcile_pebble` call, it try to start `mysqld_safe` before the data directory is initialized, causing the unit to fail to start.

## Solution

Test for data_dir initialization on the handler for `metrics-endpoint-relation-created` and only call `_reconcile_pebble` if the data_dir is initialized.


taurus-forever (approved):
Nice!

Issue #432, opened 2024-06-12
MySQL router removed from cluster set during scale up of mysql-k8s & mysql-router-k8s


carlcsaposs-canonical:
## Steps to reproduce

Unavailable https://bugs.launchpad.net/snap-openstack/+bug/2068924

Appears to be scale up of mysql-k8s & mysql-router-k8s

## Expected behavior
All units active/idle

## Actual behavior

nova-mysql-router/2 in BlockedStatus `Router was manually removed from MySQL ClusterSet. Remove & re-deploy unit`

## Versions

Juju agent: 3.4.3

mysql-k8s charm revision: 127
mysql-router-k8s charm revision: 96

## Log output
https://oil-jenkins.canonical.com/artifacts/b0040961-eb15-40f3-a89b-e49774c6f5d4/index.html

Juju debug log: 
[debug_log_openstack.txt](https://github.com/user-attachments/files/15803497/debug_log_openstack.txt)
[juju_status_openstack.txt](https://github.com/user-attachments/files/15803495/juju_status_openstack.txt)

```
unit-nova-mysql-router-0: 22:27:31 ERROR unit.nova-mysql-router/0.juju-log upgrade-version-a:44: Failed to run MySQL Shell script:
import json
import mysqlsh
import traceback

try:
    shell.connect("relation-75:***@nova-mysql-primary.openstack.svc.cluster.local:3306")
    import json

    cluster_set = dba.get_cluster_set()
    routers = cluster_set.list_routers()
    # mysqlsh objects are weird—they quack (i.e. duck typing) like standard Python objects (e.g. list,
    # dict), but do not serialize to JSON correctly.
    # Cast to str & load from JSON str before serializing
    routers = json.loads(str(routers))
    with open("/tmp/mysqlsh_output.json", "w") as file:
        json.dump(routers, file)
except mysqlsh.DBError as exception:
    error = {
        "message": str(exception),
        "code": exception.code,
        "traceback_message": "".join(traceback.format_exception(exception)),
    }
else:
    error = None
with open("/tmp/mysqlsh_error.json", "w") as file:
    json.dump(error, file)

MySQL client error 2013
MySQL Shell traceback:
Traceback (most recent call last):
  File "<string>", line 6, in <module>
mysqlsh.DBError: MySQL Error (2013): Shell.connect: Lost connection to MySQL server at 'waiting for initial communication packet', system error: 110

Traceback (most recent call last):
[...]

Issue #476, opened 2024-08-05
mysql-k8s does not create service account when related too early


gboutry:
## Steps to reproduce

1. deploy a plan with a mysql-k8s and a lot of related apps
2. wait

In single node mode, Sunbeam deploys 1 MySQL with 10 MySQL routers and 8 related apps. It is created by a terraform plan that will most likely create the integration as soon as both app are considered created by Juju. This can happen way before a charm is ready to serve relation requests, but this is independent of terraform.

Not reproducible 100%, but it does happen
## Expected behavior
Mysql-k8s should create account for all related services

## Actual behavior

Mysql-k8s does not create service accounts

## Log output

Captured from a CI run
Juju debug log:
```
2024-08-05T06:37:04.065Z [pebble] HTTP API server listening on ":38812".
2024-08-05T06:37:04.066Z [pebble] Started daemon.
2024-08-05T06:37:04.089Z [pebble] POST /v1/services 12.479793ms 202
2024-08-05T06:37:04.097Z [pebble] Service "container-agent" starting: /charm/bin/containeragent unit --data-dir /var/lib/juju --append-env "PATH=$PATH:/charm/bin" --show-log --charm-modified-version 0
2024-08-05T06:37:04.359Z [container-agent] 2024-08-05 06:37:04 INFO juju.cmd supercommand.go:56 running containerAgent [3.5.3 63d460f9ee6c7c710131961390687e7a0ab90470 gc go1.21.12]
2024-08-05T06:37:04.365Z [container-agent] starting containeragent unit command
2024-08-05T06:37:04.371Z [container-agent] containeragent unit "unit-mysql-0" start (3.5.3 [gc])
2024-08-05T06:37:04.379Z [container-agent] 2024-08-05 06:37:04 INFO juju.cmd.containeragent.unit runner.go:592 start "unit"
2024-08-05T06:37:04.379Z [container-agent] 2024-08-05 06:37:04 INFO juju.worker.upgradesteps worker.go:60 upgrade steps for 3.5.3 have already been run.
2024-08-05T06:37:04.449Z [container-agent] 2024-08-05 06:37:04 INFO juju.worker.probehttpserver server.go:157 starting http server on 127.0.0.1:65301
2024-08-05T06:37:04.476Z [container-agent] 2024-08-05 06:37:04 INFO juju.api apiclient.go:629 connection established to "wss://172.17.0.1:17070/model/32e54cee-388e-41df-8525-1817a6ff4e5a/api"
[...]

PR #489, opened 2024-08-21
DPE-4613 Test for multi-relation scale in/out

paulomach:
Issue #432 is not easily reproducible.

This PR tries to reproduce the use case, using multiple applications connected to a single cluster.


taurus-forever:
@paulomach The test passed, and looks solid.
Proposal: deploy mysql db 3 units and random(1-3) units of the router.

Mo IPs + more events inside.

tests/integration/test_multi_relations.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+@pytest.mark.group(1)
+@pytest.mark.abort_on_fail
+async def test_build_and_deploy(ops_test: OpsTest):
+    """Build the charm and deploy 1 units to ensure a cluster is formed."""
+    # Build and deploy charm from local source folder
+    db_charm = await ops_test.build_charm(".")
+
+    config = {"profile": "testing"}
```
paulomach, line 22: @taurus-forever

paulomach:
> @paulomach The test passed, and looks solid. Proposal: deploy mysql db 3 units and random(1-3) units of the router.
> 
> More IPs + more events inside.
> 
> Also, is profile=production used on purpose here?

It's on `testing`. 

I was out to replicate sunbeam deployment size + simultaneous scale in/out to catch their issue, but it may be helpful to exaggerate it here.

tests/integration/test_multi_relations.py
taurus-forever: Right... I expected to see it as line 34. Missed. Tnx!

tests/integration/test_multi_relations.py
```diff
@@ -0,0 +1,92 @@
...
+        series="jammy",
+        trust=True,
+    )
+
+    for i in range(1, SCALE_OUT + 1):
+        await ops_test.model.deploy(
+            "mysql-test-app",
+            application_name=f"app{i}",
+            num_units=1,
+            channel="latest/edge",
```
taurus-forever, line 41: auto_start_writes=false ? re: performance of GH Runners.... 
P.S. consider to add rate-limit for mysql-test-app to do NOT kill DB in such cases: https://github.com/canonical/postgresql-test-app/blob/main/config.yaml#L11-L15

So you will be able to use here `sleep_interval=1000(ms)` to rate-limit 1 insert per second from each test-app.

taurus-forever (approved):
LGTM. I would also use 3 units for router (+ some test app also for 3 units.)

One performance recomentation here due to Inserts auto-start on test-app.

shayancanonical (approved):

shayancanonical:
May be we can add a couple more `mysql-test-app` applications and `router` applications to see if we run into https://github.com/canonical/mysql-k8s-operator/issues/476. Thoughts?

PR #491, opened 2024-08-21
DPE-5209 Fix unit-initialized test may break when run too early

paulomach:
## Issue

`upgrade-relation-changed` can be triggered before workload container is ready on slow environments (looking at you GH).
Case it does, test for `unit_initialized` will fail, which can fail CI (on normal env, hook is retried until it passes).

## Solution

Test for container availability

### DIFF
--- a/src/upgrade.py
+++ b/src/upgrade.py
@@ -203,7 +203,7 @@ class MySQLK8sUpgrade(DataUpgrade):
 
         Run update status for every unit when the upgrade is completed.
         """
-        if not self.charm.unit.get_container(CONTAINER_NAME).can_connect:
+        if not self.charm.unit.get_container(CONTAINER_NAME).can_connect():
             return
         if not self.upgrade_stack and self.idle and self.charm.unit_initialized:
             self.charm._on_update_status(None)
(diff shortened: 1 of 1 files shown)


src/upgrade.py
```diff
@@ -203,6 +203,8 @@ def _on_upgrade_changed(self, _) -> None:
 
         Run update status for every unit when the upgrade is completed.
         """
+        if not self.charm.unit.get_container(CONTAINER_NAME).can_connect:
```
shayancanonical, line 206: is `can_connect` is a [function](https://ops.readthedocs.io/en/latest/#ops.Container.can_connect) and not a property. i believe you'll need `()` to cal the function

shayancanonical (changes requested):

src/upgrade.py
paulomach: absolutely right: [5811135](https://github.com/canonical/mysql-k8s-operator/pull/491/commits/5811135218eb86bd6e3cd0ef91decc8c261b93d9)

shayancanonical (approved):

taurus-forever (approved):

taurus-forever:
I am merging this as I need continue testing MySQL on ARM and it affects me. All tests passed and reviewers are happy.
Thank you for the quick fix Paulo!

PR #493, opened 2024-08-22
DPE-5210 Common credentials fixture and `exec` timeout workaround

paulomach:
## Issue

* Freeze DB test fails when waiting for credentials.
* multiple calls for same credentials during testing
* ops's `container.exec` timeout is not consistently working

## Solution

* Fixture for credentials retrieval
* workaround for `container.exec` timeout

## TODO

~~Propagate lib change back to VM~~

### DIFF
--- a/lib/charms/mysql/v0/mysql.py
+++ b/lib/charms/mysql/v0/mysql.py
@@ -1536,7 +1536,9 @@ class MySQLBase(ABC):
         )
 
         try:
-            output = self._run_mysqlsh_script("\n".join(check_cluster_metadata_commands))
+            output = self._run_mysqlsh_script(
+                "\n".join(check_cluster_metadata_commands), timeout=10
+            )
         except MySQLClientError:
             logger.warning(f"Failed to check if cluster metadata exists {from_instance=}")
             return False
--- a/lib/charms/mysql/v0/mysql.py
+++ b/lib/charms/mysql/v0/mysql.py
@@ -134,7 +134,7 @@ LIBID = "8c1428f06b1b4ec8bf98b7d980a38a8c"
 # Increment this major API version when introducing breaking changes
 LIBAPI = 0
 
-LIBPATCH = 69
+LIBPATCH = 70
 
 UNIT_TEARDOWN_LOCKNAME = "unit-teardown"
 UNIT_ADD_LOCKNAME = "unit-add"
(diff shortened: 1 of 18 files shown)


lib/charms/mysql/v0/mysql.py
```diff
@@ -1536,7 +1536,9 @@ def cluster_metadata_exists(self, from_instance: str) -> bool:
         )
 
         try:
-            output = self._run_mysqlsh_script("\n".join(check_cluster_metadata_commands))
+            output = self._run_mysqlsh_script(
+                "\n".join(check_cluster_metadata_commands), timeout=10
+            )
```
paulomach, line 1541: _Note to self_
* libpatch bump
* merge on vm

tests/integration/high_availability/test_self_healing.py
```diff
@@ -97,38 +97,41 @@ async def test_kill_db_process(
...
+    await ensure_all_units_continuous_writes_incrementing(ops_test, credentials=credentials)
 
     mysql_unit = ops_test.model.applications[mysql_application_name].units[0]
     primary = await get_primary_unit(ops_test, mysql_unit, mysql_application_name)
 
     mysql_pid = await get_process_pid(
         ops_test, primary.name, MYSQL_CONTAINER_NAME, MYSQLD_PROCESS_NAME
     )
-    assert mysql_pid > 0, "mysql process id is not positive"
+    assert (mysql_pid or -1) > 0, "mysql process id is not positive"
```
taurus-forever, line 134: 1) When can PID be missing?
2) below the error message is `is not known/positive` but here is just `not positive`. On purpose?

taurus-forever (approved):
Yeah, timeout 10 seconds should be enough for all case (c) Bill Gates.
Nice, I suspect caching credentials will speedup the all tests and finally removes pink from allure. Nice catch with container.exec() !

tests/integration/high_availability/test_self_healing.py
paulomach: this is just to quiet down the linter. the pid will be `None` when not found, causing linter messages on further use of the `mysql_pid` variable, which is very annoying

shayancanonical (approved):
awesome catches!

taurus-forever (approved):
Excellent! Thank you!

PR #496, opened 2024-08-30
Attempt to stabilize failing integration tests

shayancanonical:
## Issue
1. Even though the logrotation test kills the `log_rotate_dispatcher`, handling any event will restart the dispatcher process, leading the [following test failure](https://canonical.github.io/mysql-k8s-operator/2228/#categories/c7332569806a6cf9c4385feb46c38947/da8b6a29aef05e1f/)
2. Clear continuous writes may fail for various reasons during the self healing tests if the cluster is not healthy. For instance [this test run](https://github.com/canonical/mysql-k8s-operator/actions/runs/10625153069/job/29454851141#step:37:1037) or [this test run](https://github.com/canonical/mysql-k8s-operator/actions/runs/10633935256/job/29480518847#step:37:1077)

## Solution
1. Check that the files that need to be rotated get rotated, with room for the lack of `audit.log` file (which may be rotated but not exist yet as no db access made)
2. No clear solution. Requires investigation into separate issues

### DIFF
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -622,7 +622,7 @@ async def read_contents_from_file_in_unit(
     return contents
 
 
-async def ls_la_in_unit(
+async def ls_in_unit(
     ops_test: OpsTest,
     unit_name: str,
     directory: str,
@@ -642,18 +642,16 @@ async def ls_la_in_unit(
         a list of files returned by ls -la
     """
     return_code, output, _ = await ops_test.juju(
-        "ssh", "--container", container_name, unit_name, "ls", "-la", directory
+        "ssh", "--container", container_name, unit_name, "ls", "-1", directory
     )
     assert return_code == 0
 
-    ls_output = output.split("\n")[1:]
+    ls_output = output.split("\n")
 
     return [
         line.strip("\r")
         for line in ls_output
-        if len(line.strip()) > 0
-        and line.split()[-1] not in exclude_files
-        and line.split()[-1] not in [".", ".."]
+        if len(line.strip()) > 0 and line.strip() not in exclude_files
     ]
 
 
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -356,6 +356,7 @@ async def get_process_pid(
         unit_name: The name of the unit
         container_name: The name of the container in the unit
         process: The process name to search for
+        full_match: Whether to fully match the process name
 
     Returns:
         A integer for the process id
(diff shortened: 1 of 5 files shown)


tests/integration/helpers.py
```diff
@@ -645,7 +651,9 @@ async def ls_la_in_unit(
     return [
         line.strip("\r")
         for line in ls_output
-        if len(line.strip()) > 0 and line.split()[-1] not in [".", ".."]
+        if len(line.strip()) > 0
+        and line.split()[-1] not in exclude_files
+        and line.split()[-1] not in [".", ".."]
```
paulomach, line 656: 1. You can probably ditch the `[".", ".."]` exclude by not using `-a` from `ls`.
2. Do we need the long listing (`-l`) output for something? If the idea is just to have one entry per line, we can just use `ls -1` instead. EDIT: found why later in the test. One could simplify the test to not differentiate dir/files

tests/integration/high_availability/conftest.py
```diff
@@ -33,8 +34,10 @@ async def continuous_writes(ops_test: OpsTest) -> None:
 
     yield
 
-    logger.info("Clearing continuous writes")
-    await juju_.run_action(application_unit, "clear-continuous-writes")
+    for attempt in Retrying(stop=stop_after_attempt(10), wait=wait_fixed(30)):
+        with attempt:
+            logger.info("Clearing continuous writes")
+            await juju_.run_action(application_unit, "clear-continuous-writes")
```
paulomach, line 40: Got any insight why this fails on the test app?

paulomach (approved):
Minor comments, otherwise excited to see tests going on 1st run. Good stuff

tests/integration/high_availability/conftest.py
shayancanonical: There are a few reasons (like [this](https://github.com/canonical/mysql-k8s-operator/actions/runs/10625153069/job/29454851141#step:37:1037) where a relation user does not exist for the test app to clear continuous writes and [this](https://github.com/canonical/mysql-k8s-operator/actions/runs/10633935256/job/29480518847#step:37:1077) where the database is in `--read-only` mode)

I am starting to lean toward's Alex's caution - may be we shouldnt retry clearing continuous writes and instead focus on the root cause which leads to the relation user not being available or that which leads to the database being in read-only mode

tests/integration/helpers.py
shayancanonical: great suggestion. addressed in [022a4cc](https://github.com/canonical/mysql-k8s-operator/pull/496/commits/022a4cc10ae7199d92283c84913f0531bc587385)

tests/integration/high_availability/conftest.py
paulomach: In any case, it seems more like a test app issue.
For that we need to either treat error or fail when it should there, validating test-app status in test.
imo, we should not block this though

Issue #459, opened 2024-07-10
MySQL Error (1045): Shell.connect: Access denied


cjohnston1158:
## Steps to reproduce
This is a deployment of Sunbeam with MAAS. I don't see the issue 100% of the time, and I'm not sure yet any specific steps to reproduce.

## Expected behavior
No error

## Actual behavior
```
juju status -m openstack placement-mysql
Model      Controller            Cloud/Region                Version  SLA          Timestamp
openstack  sunbeam01-controller  sunbeam-microk8s/localhost  3.4.4    unsupported  12:41:42Z

SAAS       Status  Store  URL
microceph  active  local  admin/openstack-machines.microceph

App              Version                  Status   Scale  Charm      Channel   Rev  Address         Exposed  Message
placement-mysql  8.0.37-0ubuntu0.22.04.3  waiting      3  mysql-k8s  8.0/edge  160  10.152.183.211  no       installing agent

Unit                Workload     Agent      Address      Ports  Message
placement-mysql/0   maintenance  idle       10.1.250.83         joining the cluster
placement-mysql/1   waiting      idle       10.1.230.20         waiting to get cluster primary from peers
placement-mysql/2*  maintenance  executing  10.1.51.149         Unable to get member state
```

```
2024-07-10T09:10:41.484Z [container-agent] 2024-07-10 09:10:41 ERROR juju-log Failed to flush [<MySQLTextLogs.ERROR: 'ERROR LOGS'>, <MySQLTextLogs.GENERAL: 'GENERAL LOGS'>, <MySQLTextLogs.SLOW: 'SLOW LOGS'>] logs.
2024-07-10T09:10:41.484Z [container-agent] Traceback (most recent call last):
2024-07-10T09:10:41.484Z [container-agent]   File "/var/lib/juju/agents/unit-placement-mysql-2/charm/src/mysql_k8s_helpers.py", line 611, in _run_mysqlsh_script
2024-07-10T09:10:41.484Z [container-agent]     stdout, _ = process.wait_output()
2024-07-10T09:10:41.484Z [container-agent]   File "/var/lib/juju/agents/unit-placement-mysql-2/charm/venv/ops/pebble.py", line 1635, in wait_output
2024-07-10T09:10:41.484Z [container-agent]     raise ExecError[AnyStr](self._command, exit_code, out_value, err_value)
[...]

PR #501, opened 2024-09-06
[DPE-4861] Avoid ambiguous service selector when multiple mysql apps in a model have the same cluster-name

shayancanonical:
## Issue
When 2 mysql applications exist in the same model with the same cluster-name, it is possible that traffic for one cluster gets routed to the other cluster

Example: https://github.com/canonical/mysql-k8s-operator/actions/runs/10730746555/job/29759894942#step:37:1051
Possibly related: https://github.com/canonical/mysql-k8s-operator/issues/459

This is because our selector for service endpoint is `cluster-name` and `role`

## Solution
Add `application-name` as a label to pods, and use this as a selector for the service endpoint as well

### DIFF
--- a/src/k8s_helpers.py
+++ b/src/k8s_helpers.py
@@ -54,7 +54,11 @@ class KubernetesHelpers:
             roles: List of roles to append on the service name
         """
         for role in roles:
-            selector = {"cluster-name": self.cluster_name, "role": role}
+            selector = {
+                "cluster-name": self.cluster_name,
+                "application_name": self.app_name,
+                "role": role,
+            }
             service_name = f"{self.app_name}-{role}"
             pod0 = self.client.get(
                 res=Pod,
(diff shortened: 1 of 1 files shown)


taurus-forever (approved):
Nice catch!

I do not see, how it can fix https://github.com/canonical/mysql-k8s-operator/issues/459 , however your logic looks conceptually correct.

taurus-forever:
Q: who will update K8s selector if user deploy old revision and refresh to the new charm revision? ;-)

shayancanonical:
Very wishful thinking, but the reporter for issue 459 confirmed that there are multiple deployed MySQL applications when sunbeam is deployed. My thoughts were that it may be possible that there may be two applications with the same generated `cluster-name`s (confirmed with Guillaume from sunbeam that `cluster-name` is not explicitly configured). If this is the case, traffic for one cluster's primary may be redirected to another cluster's primary. However, I now realize that issue 459 happens when flushing logs - more investigation required

On `update-status`, if the cluster is healthy, we `update-endpoints` which results in the correct patch of pod labels. When a charm is upgraded, the next update status should correct k8s selector. Furthermore, upon `complete-upgrade` (or rather `pebble-ready` after restart of pod) of `mysql-k8s/1`, we call `set_cluster_primary` to set unit 1 as the primary before upgrading unit 0. Setting this primary will update the pod labels correctly, so the k8s selector will work right after upgrading the charm to the new revision

paulomach (approved):

PR #508, opened 2024-09-17
[DPE-4861] Add test to ensure correct k8s endpoints created for clusters with the same name

shayancanonical:
## Issue
We are missing an integration test for https://github.com/canonical/mysql-k8s-operator/pull/501

## Solution
Add integration test

### DIFF
--- a/tests/integration/high_availability/test_k8s_endpoints.py
+++ b/tests/integration/high_availability/test_k8s_endpoints.py
@@ -1,0 +1,114 @@
+#!/usr/bin/env python3
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+import logging
+
+import pytest
+from pytest_operator.plugin import OpsTest
+
+from ..helpers import get_unit_address
+from .high_availability_helpers import (
+    deploy_and_scale_application,
+    deploy_and_scale_mysql,
+    get_endpoint_addresses,
+    relate_mysql_and_application,
+)
+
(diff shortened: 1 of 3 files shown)


paulomach (approved):
:heart_eyes:

tests/integration/high_availability/test_k8s_endpoints.py
```diff
@@ -0,0 +1,114 @@
...
+MYSQL_CLUSTER_ONE = "mysql1"
+MYSQL_CLUSTER_TWO = "mysql2"
+MYSQL_CLUSTER_NAME = "test_cluster"
+TEST_APP_ONE = "mysql-test-app1"
+TEST_APP_TWO = "mysql-test-app2"
+
+
+@pytest.mark.group(1)
+@pytest.mark.abort_on_fail
+async def test_labeling_of_k8s_endpoints(ops_test: OpsTest):
```
paulomach, line 29: the one thing to dispute here is the test being under ha tests, but I guess it's due usage of helpers, and not about being ha. we may need to shuffle some of the helpers for organization sake

taurus-forever (approved):
Thank you!

tests/integration/high_availability/test_k8s_endpoints.py
shayancanonical: agreed, let's rearrange furniture in a separate PR allowing a better classification of integration tests

PR #522, opened 2024-10-11
DPE-5627 CC006 compliance

paulomach:


taurus-forever (approved):
:+1:

paulomach:
> if user pins charm revision, does terraform select correct OCI resource rev?

It's handled by juju, so yes. Unless it's overridden

paulomach:
> does cc006 require any versioning of the module?

https://github.com/canonical/mysql-operator/pull/540#issuecomment-2413795508

PR #539, opened 2024-12-10
[DPE-5588] Check against invalid arch

sinclert-canonical:
This PR introduces the usage of the newly created architecture lib (see [#560](https://github.com/canonical/mysql-operator/pull/560)) into the charm itself.

The shape of the integration tests, as well as the helper used to get the built charm in the architecture that mismatches the one of the host itself, have been adapted from the one Lucas created for PostgreSQL-k8s (see [code](https://github.com/canonical/postgresql-k8s-operator/blob/70a96ce33eb0de054b9297e3bd558f6ece54c0b0/tests/integration/test_wrong_arch.py)).

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -4,6 +4,12 @@
 
 """Charm for MySQL."""
 
+from charms.mysql.v0.architecture import WrongArchitectureWarningCharm, is_wrong_architecture
+from ops.main import main
+
+if is_wrong_architecture() and __name__ == "__main__":
+    main(WrongArchitectureWarningCharm)
+
 import logging
 import random
 from socket import getfqdn
@@ -44,7 +50,6 @@ from charms.tempo_coordinator_k8s.v0.charm_tracing import trace_charm
 from charms.tempo_coordinator_k8s.v0.tracing import TracingEndpointRequirer
 from ops import EventBase, RelationBrokenEvent, RelationCreatedEvent
 from ops.charm import RelationChangedEvent, UpdateStatusEvent
-from ops.main import main
 from ops.model import (
     ActiveStatus,
     BlockedStatus,
--- a/lib/charms/mysql/v0/architecture.py
+++ b/lib/charms/mysql/v0/architecture.py
@@ -1,0 +1,93 @@
+# Copyright 2024 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Library to provide hardware architecture checks for VMs and K8s charms.
+
+The WrongArchitectureWarningCharm class is designed to be used alongside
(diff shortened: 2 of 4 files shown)


src/charm.py
```diff
@@ -10,6 +10,18 @@
 from time import sleep
 from typing import Optional
 
+try:
+    # Library depends on 'cryptography', a platform-specific dependency.
+    # Importing it will fail on wrong architecture.
+    from charms.mysql.v0.tls import MySQLTLS
```
sinclert-canonical, line 16: This is the first import that I found fails due to the architecture it gets built on. It fails because it relies, indirectly, on the `cryptography` package. As an alternative, we could try to import `cryptography` directly, although Ruff would complaint that the import would not be used anywhere.

Let me know what you think.

sinclert-canonical: > I would check is_wrong_architecture before any imports

You mean even before the Standard library imports? Yeah, we can move it further up.

> I don't think we should raise since that will obscure the blocked status

The `raise` clause is there for cases where the import fails for any other reason (i.e. forget to change its name after a refactor, deploying in the wrong architecture but not actually executing the charm code, etc).

For the case that concern us, the `WrongArchitectureWarningCharm` class already exits with code `0` (see [code](https://github.com/canonical/mysql-operator/blob/f1cce241dc11eed4ee31aba2a3cfb638c4c03edb/lib/charms/mysql/v0/architecture.py#L60)), so it will never re-raise the exception. I can move the `raise` clause to an `else` block if that makes it easier to read.

sinclert-canonical: > I would check is_wrong_architecture before any imports

You mean even before the Standard library imports? Yeah, we can move it further up.

> I don't think we should raise since that will obscure the blocked status

The `raise` clause is there for cases where the import fails for any other reason (i.e. forget to change its name after a refactor, deploying in the wrong architecture but not actually executing the charm code, etc).

For the case that concern us, the `WrongArchitectureWarningCharm` class already exits with code `0` (see [code](https://github.com/canonical/mysql-operator/blob/f1cce241dc11eed4ee31aba2a3cfb638c4c03edb/lib/charms/mysql/v0/architecture.py#L60)), so it will never re-raise the exception. I can move the `raise` clause to an `else` block if that makes it easier.

sinclert-canonical: If you want to get rid of the `try-except` block (something I would also love to), we would need to tweak the `is_wrong_architecture` lib helper. Turns out unit tests do not set up the `CHARM_DIR` env. variable, and we assume that whenever we run that helper function it will always be available (see [code](https://github.com/canonical/mysql-operator/blob/f1cce241dc11eed4ee31aba2a3cfb638c4c03edb/lib/charms/mysql/v0/architecture.py#L65)).

Just noticed: doing what the PostgreSQL folks do in their repo would have worked (see [code](https://github.com/canonical/postgresql-k8s-operator/blob/70a96ce33eb0de054b9297e3bd558f6ece54c0b0/src/arch_utils.py#L29)) 😅 

```python
manifest_file_path = f"os.environ.get('CHARM_DIR')/manifest.yaml"
```

sinclert-canonical: Let me tweak the architecture lib to make it more resilient. No repo is using it so far, so I would say it is OK.

sinclert-canonical:
> still have concerns about rollbacks & how this could makes things worse than if we continue to raise uncaught exception ([#539 (review)](https://github.com/canonical/mysql-k8s-operator/pull/539#pullrequestreview-2494754414))

Not sure what you mean by that. Mind bringing this up during today's Technical Sync?

sinclert-canonical:
Rebased after PR [#564](https://github.com/canonical/mysql-operator/pull/564) was merged.

PR #553, opened 2025-01-14
fix: DPE-6137 insecure password calls removal

paulomach:
Port of VM [PR#579](https://github.com/canonical/mysql-operator/pull/579), plus:

* `mysql.py` shared lib update that:
  * disable general query log
  * changes log level for destructive actions (i.e. `s/debug/info`)


src/mysql_k8s_helpers.py
```diff
@@ -618,21 +638,21 @@ def _run_mysqlsh_script(
...
         try:
-            process = self.container.exec(cmd)
+            process = self.container.exec(cmd, stdin=password)
             stdout, _ = process.wait_output()
-            return stdout
-        except ExecError:
-            raise MySQLClientError
-        except ChangeError:
-            raise MySQLClientError
+            return stdout.split("###")[1].strip()
```
sinclert-canonical, line 643: I understand we are removing a `###` prefix from the output, with `output` being similar to [this example](https://github.com/canonical/mysql-k8s-operator/blob/5d41681dddca41d4f2032f4d33b6930ef41cd366/tests/unit/test_mysql_k8s_helpers.py#L254) from the test suite. However, should we use Python 3.9+ [`removeprefix`](https://docs.python.org/3/library/stdtypes.html#str.removeprefix) function to better express the intent? 🤔

paulomach: If the `###` were the only thing being in the stdout yes. The point being that it can exist other printouts (warning messages, startup banners) by mysqlcli/mysqlsh that we do not control, so we print and split to get rid of unknown garbage.

sinclert-canonical: Thanks for the context. Maybe worth including that scenario within the tests, something like:

```python
return_value = "blahblah###stdout"
```

src/mysql_k8s_helpers.py
```diff
@@ -648,23 +668,26 @@ def _run_mysqlcli_script(
...
-        if password:
-            # password is needed after user
-            command.append(f"--password={password}")
 
         try:
-            process = self.container.exec(command, timeout=timeout)
+            if password:
+                # password is needed after user
+                command.insert(3, "-p")
+                process = self.container.exec(command, timeout=timeout, stdin=password)
```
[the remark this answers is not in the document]
paulomach: Yes, mysqlcli will read password from console input, not stdin. Hence using pexpect, that will wrap the call in a pseudo tty and emulate console input.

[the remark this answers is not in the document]
paulomach: Pebble is doing something else. You can try this, with a running local pebble and mysql:

```python
import subprocess
import pexpect
from ops.pebble import Client

cmd = [
    "mysql",
    "-u",
    "root",
    "-S",
    "/var/snap/charmed-mysql/common/var/run/mysqld/mysqld.sock",
    "-N",
    "-B",
    "-p",
    "-e",
    "Select user,host from mysql.user",
]

c = cmd.copy()
c[-1] = f"\"{c[-1]}\""
child = pexpect.spawnu(" ".join(c))
child.expect("Enter password:")
child.sendline("newpass")
stdout = child.readlines()

print("expect")
re = [line.strip().split() for line in stdout[1:]]


print("\nsubprocess")
stdout = subprocess.check_output(cmd, text=True, input="newpass")
rs = [line.split() for line in stdout.strip().split("\n")]


print("\npebble")
client = Client("/tmp/.pebble.socket")
p = client.exec(cmd, stdin="newpass")
stdout, _ = p.wait_output()
rp = [line.split("\t") for line in  stdout.strip().split("\n")]


print(rp == re)
print(re == rs)
```

Will output:
```shell
(venv) root@noble:~# python test.py
expect

subprocess
Enter password:  # <-- running with subprocess still requires a console input, blocking execution

pebble
True
True
```

src/mysql_k8s_helpers.py
paulomach: c0d8b6e9

sinclert-canonical (approved):

paulomach:
upgrade test for arm consistently failing on main and other branches. Merging as is to unblock other PRs

PR #560, opened 2025-01-21
[DPE-4375] Add cluster manual re-join handler

sinclert-canonical:
This PR adds logic to manually re-join a MySQL replica that has gone `OFFLINE`, to the cluster, whenever MySQL 8.0.21+ [auto re-join](https://dev.mysql.com/doc/refman/8.0/en/group-replication-responses-failure-rejoin.html) attempts have been exhausted.

### Description

There are edge cases where a MySQL instance that has lost connection to the cluster it originally belong to (i.e. `OFFLINE`), exhaust its automatic retries (by default 3 retries, with 5 mins between each, starting with MySQL 8.0.21). For those cases, a manual re-join should be performed.

---

Depends on https://github.com/canonical/mysql-k8s-operator/pull/563.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -764,7 +764,7 @@ class MySQLOperatorCharm(MySQLCharmBase, TypedCharmBase[CharmConfig]):
             logger.exception("Failed to initialize primary")
             raise
 
-    def _handle_potential_cluster_crash_scenario(self) -> bool:
+    def _handle_potential_cluster_crash_scenario(self) -> bool:  # noqa: C901
         """Handle potential full cluster crash scenarios.
 
         Returns:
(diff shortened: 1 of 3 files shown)


src/charm.py
```diff
@@ -832,10 +832,38 @@ def _handle_potential_cluster_crash_scenario(self) -> bool:
                     else:
                         self.unit.status = BlockedStatus("failed to recover cluster.")
 
+            elif self._mysql.is_cluster_auto_rejoin_ongoing():
+                logger.info("Cluster auto-rejoin attempts are still ongoing.")
+
+            elif not self._mysql.is_cluster_auto_rejoin_ongoing():
+                logger.info("Cluster auto-rejoin attempts are exhausted. Attempting manual rejoin")
+                self._execute_manual_rejoin()
```
sinclert-canonical, line 840: Both these `elif` clauses had a `not self.is_leader()` check before. However, Paulo confirmed that there are scenarios where a replica can end up as a Juju leader, and the manual re-join logic must affect those too.

src/charm.py
```diff
@@ -832,10 +832,41 @@ def _handle_potential_cluster_crash_scenario(self) -> bool:
...
```
sinclert-canonical, line 840: Both these `elif` clauses had a `not self.unit.is_leader()` check before. However, Paulo confirmed that there are scenarios where a replica can end up as a Juju leader, and the manual re-join logic must affect those too.

sinclert-canonical:
PR waiting on https://github.com/canonical/mysql-k8s-operator/pull/553 to be merged in order to update `mysql` charm lib.

paulomach (approved):

PR #570, opened 2025-01-30
[DPE-5908] Support old bucket endpoints format

sinclert-canonical:
This PR ports PostgreSQL operator `_construct_endpoint` helper function ([see code](https://github.com/canonical/postgresql-operator/blob/c5c916d41d00849da5ae657cba4c97c46cb3caad/src/backups.py#L229-L249)) , in order to support the specification of the old S3 endpoints format (i.e. `https://s3.amazonaws.com`), stated in the [charm docs](https://charmhub.io/mysql-k8s/docs/h-configure-s3-aws).

I decided to port the helper, instead of updating the docs, in order to make MySQL operator support as similar to PostgreSQL operator as possible. If we eventually decide to stop supporting the old S3 endpoints format, we will do so across our repositories, and not across a sub-set of them.

### DIFF
--- a/lib/charms/mysql/v0/s3_helpers.py
+++ b/lib/charms/mysql/v0/s3_helpers.py
@@ -21,6 +21,7 @@ import time
 from typing import Dict, List, Tuple
 
 import boto3
+import botocore
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


paulomach (approved):
lgtm

shayancanonical (approved):

PR #571, opened 2025-02-03
fix: DPE-6521 Add missing check and wait for pebble calls

paulomach:
## Issue

1. Call to pebble fails due to missing readiness check
2. Call to pebble exec without a `wait` for the process (call is not executed)

## Solution

Add said calls.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -144,3 +144,5 @@ Makefile
 
 # local pyright settings
 pyrightconfig.json
+
+wt*/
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -144,3 +144,5 @@ Makefile
 
 # local pyright settings
 pyrightconfig.json
+
+wt*/
```
paulomach, line 148: included to ignore worktree dirs

shayancanonical (approved):
ay ay ay, what a slip up. great catch!

PR #573, opened 2025-02-04
Use `charmcraft test` & concierge

carlcsaposs-canonical:
Ported from https://github.com/canonical/mysql-router-k8s-operator/pull/379

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -32,7 +32,7 @@ jobs:
     uses: ./.github/workflows/ci.yaml
     secrets: inherit
     permissions:
-      contents: write  # Needed for Allure Report beta
+      contents: write  # Needed for Allure Report
 
   release:
     name: Release charm


taurus-forever (approved):
I have feeling we need an intro to the new design. 
LGTM, question about safety of env variables on non-trusted hosts.
Curious about possible simplifications of `test_healing_*.py`...
The final decission on @paulomach here.

spread.yaml
```diff
@@ -0,0 +1,134 @@
...
+  tests/spread/:
+    summary: Spread tests
+
+path: /root/spread_project
+
+kill-timeout: 3h
+environment:
+  PATH: $PATH:$(pipx environment --value PIPX_BIN_DIR)
+  CONCIERGE_JUJU_CHANNEL/juju36: 3.6/stable
+  CONCIERGE_JUJU_CHANNEL/juju29: 2.9/stable
```
taurus-forever, line 109: Should you introduce CHARM_UBUNTU_BASE: 22.04 (for future 24.04) ?

sinclert-canonical (commented):
Same comments as in https://github.com/canonical/mysql-router-operator/pull/210.

PR #578, opened 2025-02-10
[MISC] Refresh all charm libs

sinclert-canonical:
This PR refreshes all charm libs that have fallen out-of-date.

I needed to bump `lighkube` minimum version to satisfy the new `cos >= 0.0.50` constraint.


sinclert-canonical:
**Disclaimer:**

I needed to revert MySQL libraries `architecture.py` and `mysql.py` to the state in which they are in Charmhub. Otherwise, the check-lib CI step crashed because contents did not match (see [failing CI job](https://github.com/canonical/mysql-k8s-operator/actions/runs/13236998441/job/36946298804)).

paulomach:
> > **Disclaimer:**
> > I needed to revert MySQL libraries `architecture.py` and `mysql.py` to the state in which they are in Charmhub. Otherwise, the check-lib CI step crashed because contents did not match (see [failing CI job](https://github.com/canonical/mysql-k8s-operator/actions/runs/13236998441/job/36946298804)).
> 
> I think we need to publish a new libpatch on mysql vm instead

Yes. 0.83 was updated yesterday on mysql-vm (and charmhub), so you should be able to update this with the correct one

sinclert-canonical:
Thanks @paulomach!

Just rebased my branch from `main` and remove the reverting of the `mysql` charm lib version.

paulomach (approved):
Thanks!

taurus-forever (approved):
Thank you!!!

PR #593, opened 2025-02-21
[DPE-6488] Port over mysql charmlib changes from vm operator related to unit initialized

shayancanonical:
Porting over changes from https://github.com/canonical/mysql-operator/pull/615 in order to test that the changes are compatible with the K8s charm as well


taurus-forever (approved):
LGTM, the final ACK is on @paulomach .

src/charm.py
```diff
@@ -890,7 +889,7 @@ def _handle_potential_cluster_crash_scenario(self) -> bool:  # noqa: C901
 
             if self._mysql.is_cluster_auto_rejoin_ongoing():
                 logger.info("Cluster auto-rejoin attempts are still ongoing.")
-            else:
+            elif all_states != {"offline"}:
```
sinclert-canonical, line 892: Please double-check that this is needed, as discussed privately.

sinclert-canonical (approved):

src/charm.py
shayancanonical: double-checked: the redundant check is not necessary. the error trace i was seeing only happens in `update-status` when one unit is offline and other units are shutting down (but the charm has not realized this yet)

reverted in c737514338665dfbe282c621864281e2bf4697d7

paulomach (approved):
good-to-go

PR #594, opened 2025-02-24
[DPE-6651] Update rock tag for new builds with updated percona packages

shayancanonical:
## Issue
We have new rock builds that include new versions of percona packages

## Solution
Update rock image


sinclert-canonical (approved):

taurus-forever (approved):

paulomach (approved):
remarkable

PR #600, opened 2025-03-07
DPE-6487 Pitr

paulomach:
Port of VM [PR#621](https://github.com/canonical/mysql-operator/pull/621)


lib/charms/mysql/v0/backups.py
```diff
@@ -85,7 +93,11 @@ def is_unit_blocked(self) -> bool:
 from ops.jujuversion import JujuVersion
 from ops.model import BlockedStatus, MaintenanceStatus
 
-from constants import MYSQL_DATA_DIR
+from constants import (
+    MYSQL_DATA_DIR,
+    SERVER_CONFIG_PASSWORD_KEY,
+    SERVER_CONFIG_USERNAME,
+)
```
sinclert-canonical, line 100: I know this is not the point of the PR, but changes like this one does not move us in the right direction.

The MySQL charm libs (_backups_, _mysql_, _helpers_...) will never behave like real libraries, until we force ourselves to maintain good hygiene about what we import from where. A "library" that imports application code is no "library".

lib/charms/mysql/v0/backups.py
```diff
@@ -519,14 +574,28 @@ def _on_restore(self, event: ActionEvent) -> None:
...
+            if not success:
+                logger.error(f"Restore failed: {error_message}")
+                event.fail(error_message)
+                self.charm.unit.status = BlockedStatus(error_message)
+                return
+
+        self.charm.app_peer_data.update({
+            "s3-block-message": MOVE_RESTORED_CLUSTER_TO_ANOTHER_S3_REPOSITORY_ERROR,
+            "binlogs-collecting": "",
+        })
```
sinclert-canonical, line 595: Should we check that the instance calling this method is the Juju leader? Otherwise, it won't be able to write on the Application databag.

lib/charms/mysql/v0/backups.py
```diff
@@ -674,3 +761,130 @@ def _post_restore(self) -> Tuple[bool, str]:
...
+        except Exception as e:
+            logger.error(e)
+            logger.error(
+                "Exception is occurred when trying to stop binlogs collecting after S3 relation depart. It may be a leader departure"
+            )
+
+    def get_binlogs_collector_config(self) -> Dict[str, str]:
+        """Update binlogs collector service config file.
+
+        Returns: whether this operation was successful.
```
sinclert-canonical, line 842: I think this doc-string fell out of date. It should reflect the dictionary structure being returned.

lib/charms/mysql/v0/backups.py
```diff
@@ -674,3 +761,130 @@ def _post_restore(self) -> Tuple[bool, str]:
...
+            logger.error(
+                f"Cannot get binlogs collector config: Missing S3 parameters: {missing_parameters}"
+            )
+            return {}
+
+        bucket_url = (
+            f"{s3_parameters['bucket']}/{s3_parameters['path']}binlogs"
+            if not s3_parameters["path"] or s3_parameters["path"][-1] == "/"
+            else f"{s3_parameters['bucket']}/{s3_parameters['path']}/binlogs"
+        )
```
sinclert-canonical, line 862: There is a better way to deal with optional trailing `/` character: to use [rstrip](https://docs.python.org/3/library/stdtypes.html#str.rstrip).

Feel free to in-line the following unfolded example:
```python
binlogs_path = s3_parameters.get("path", "").rstrip("/")
bucket_name = s3_parameters.get("bucket", "")
bucket_url = f"{bucket_name}/{binlogs_path}/binlogs"
```

lib/charms/mysql/v0/backups.py
```diff
@@ -674,3 +761,130 @@ def _post_restore(self) -> Tuple[bool, str]:
...
+        if not re.match(
+            r"^\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}$",
+            timestamp,
+        ):
+            return False
+        try:
+            self._parse_mysql_timestamp(timestamp)
+            return True
+        except ValueError:
+            return False
```
sinclert-canonical, line 887: What's wrong with just relying on the `try-except` block? Why do we need to regex-match? 🤔

lib/charms/mysql/v0/mysql.py
```diff
@@ -3128,6 +3140,62 @@ def restore_backup(
...
+            s3_parameters: S3 relation parameters.
+            restore_to_time: the MySQL timestamp to restore to or keyword `latest`.
+            user: the user with which to execute the commands.
+            group: the group with which to execute the commands.
+        """
+        bucket_url = (
+            f"{s3_parameters['bucket']}/{s3_parameters['path']}binlogs"
+            if s3_parameters["path"][-1] == "/"
+            else f"{s3_parameters['bucket']}/{s3_parameters['path']}/binlogs"
+        )
```
sinclert-canonical, line 3168: Same as a previous comment, I think it is better to use [rstrip](https://docs.python.org/3/library/stdtypes.html#str.rstrip).

lib/charms/mysql/v0/s3_helpers.py
```diff
@@ -75,44 +81,103 @@ def upload_content_to_s3(content: str, content_path: str, s3_parameters: Dict) -
...
         )
         return False
-    finally:
-        if ca_file:
-            ca_file.close()
 
     return True
 
 
+def _read_content_from_s3(content_path: str, s3_parameters: dict) -> str | None:
```
sinclert-canonical, line 122: I would probably make this function public, mimicking its `upload_contents_to_s3` counterpart.

lib/charms/mysql/v0/s3_helpers.py
```diff
@@ -75,44 +81,103 @@ def upload_content_to_s3(content: str, content_path: str, s3_parameters: Dict) -
...
+                ca_file.write(ca.encode())
+                ca_file.flush()
+
+            s3 = session.resource(
+                "s3",
+                endpoint_url=_construct_endpoint(s3_parameters),
+                verify=ca_file.name if ca_file else True,
+            )
+
+            bucket = s3.Bucket(s3_parameters["bucket"])
```
sinclert-canonical, line 157: Are not these the same ~20 LOC as in the `upload_contents_to_s3` [function](https://github.com/canonical/mysql-k8s-operator/blob/cbcab2a7e82f634121f76752411b2198141a98c4/lib/charms/mysql/v0/s3_helpers.py#L72-L119)? Maybe worth extracting common logic into a private function.

src/mysql_k8s_helpers.py
```diff
@@ -931,3 +932,56 @@ def _file_exists(self, path: str) -> bool:
...
+        return True
+
+    def get_cluster_members(self) -> list[str]:
+        """Get cluster members in MySQL MEMBER_HOST format.
+
+        Returns: list of cluster members in MySQL MEMBER_HOST format.
+        """
+        return [self.charm.get_unit_address(self.charm.unit)] + [
+            self.charm.get_unit_address(unit) for unit in self.charm.peers.units
+        ]
```
sinclert-canonical, line 987: There may be a more pythonic way to write this:

```python
return [
    self.charm.get_unit_address(self.charm.unit),
    *(self.charm.get_unit_address(peer_unit) for peer_unit in self.charm.peers.units),
]
```

sinclert-canonical (commented):
LGTM overall.

lib/charms/mysql/v0/backups.py
Zvirovyi: `_pre_restore_checks` will check on `self.charm.app.planned_units()` to ensure there is only a single unit therefore a leader.

Zvirovyi:
@paulomach Should I intervene? If so, how to do it with your PR?

lib/charms/mysql/v0/backups.py
paulomach: good catch. @Zvirovyi  see any issues with :

```suggestion
        try:
            datetime.datetime.strptime(timestamp, "%Y-%m-%d %H:%M:%S")
            return True
        except ValueError:
            return False
```

lib/charms/mysql/v0/s3_helpers.py
paulomach: There's no current public usage, so I rather keep it private

lib/charms/mysql/v0/backups.py
Zvirovyi: With the suggested approach, it will accept values such as `2025-2-02 1:10:10` in Python, while `time.Parse` in [mysql-pitr-helper](https://github.com/canonical/mysql-pitr-helper/blob/ed858df5c145b003c9d24223d44b6ea9c7d67888/recoverer/recoverer.go#L194) (Go) will return an error ([example](https://go.dev/play/p/uvtfFi1CLrR)). My advice is to keep regex to enforce a strict format across different projects/languages regardless of their own time-parsing implementation.

On the other hand, we can loosen the check on the mysql-pitr-helper side, but it will require testing to ensure full compatibility and I don't see it viable.

sinclert-canonical: Thanks for clarifying.

I would probably add a test failing if the string date does not follow the plugin expected format, or a comment stating why the regex is there (i.e. required by the Go plugin 2-3 abstraction layers deeper). Otherwise, I am afraid this snippet may be mistakenly removed in the future.

paulomach: I'll add a comment

lib/charms/mysql/v0/backups.py
sinclert-canonical: I see feedback addressed in commit https://github.com/canonical/mysql-k8s-operator/pull/600/commits/936628588741d428d92d606e2c4a801ffce0e0f1.

lib/charms/mysql/v0/backups.py
sinclert-canonical: I see feedback addressed in commit https://github.com/canonical/mysql-k8s-operator/pull/600/commits/936628588741d428d92d606e2c4a801ffce0e0f1.

lib/charms/mysql/v0/mysql.py
sinclert-canonical: I see feedback addressed in commit https://github.com/canonical/mysql-k8s-operator/pull/600/commits/936628588741d428d92d606e2c4a801ffce0e0f1.

lib/charms/mysql/v0/s3_helpers.py
sinclert-canonical: I see feedback addressed in commit https://github.com/canonical/mysql-k8s-operator/pull/600/commits/936628588741d428d92d606e2c4a801ffce0e0f1.

sinclert-canonical (approved):
With most of my feedback addressed, this is good for me to merge.

Fantastic work everyone! 🚀

taurus-forever (approved):
Incredible! Well done @Zvirovyi !

Issue #614, opened 2025-05-29
Failed on MySQLGetClusterPrimaryAddressError


jeffreychang911:
## Steps to reproduce
1. bootstrap a juju 3.6.5 controller on canonical k8s 1.32
2. deploy the bundle [here](https://radosgw.ps5.canonical.com/swift/v1/AUTH_f7786960f7b042b9b9e84bcf30e51d75/charm-qa/production/148054/generated-bundle.yaml)
3. check juju status

## Expected behavior
mysql-k8s should be in active/idle state.

## Actual behavior
App          Version  Status   Scale  Charm        Channel         Rev  Address         Exposed  Message
istio-pilot           blocked      1  istio-pilot  1.24/stable    1359  10.152.183.221  no       Execution handled 2 errors.  See logs for details.
mongodb-k8s           active       1  mongodb-k8s  6/stable         61  10.152.183.57   no       
neighbor              error        1  mysql-k8s    8.0/stable      240  10.152.183.38   no       hook failed: "leader-elected"
target                blocked      1  osm-ro       latest/stable     1                  no       Need kafka, mongodb relations
vault-k8s             active       1  vault-k8s    1.16/stable     323  10.152.183.238  no      

## Versions
Juju CLI: 3,6,5
Juju agent: 3.6.5
mysql-k8s charm revision: 8.0/stable rev 240
canonical-k8s: 1.32

## Log output
Juju debug log: 
2025-05-28T14:08:54.953Z [container-agent] 2025-05-28 14:08:54 DEBUG juju-log Getting cluster primary member's address
2025-05-28T14:08:54.978Z [container-agent] 2025-05-28 14:08:54 DEBUG juju.worker.uniter pebblenotices.go:158 container "mysql": ignoring change-update notice, kind exec
2025-05-28T14:08:55.279Z [container-agent] 2025-05-28 14:08:55 DEBUG juju.worker.uniter pebblenotices.go:158 container "mysql": ignoring change-update notice, kind exec
2025-05-28T14:08:55.285Z [container-agent] 2025-05-28 14:08:55 DEBUG jujuc server.go:303 running hook tool "juju-log" for neighbor/0-leader-elected-89646a056b701e2a4675a27eb9aeb316
2025-05-28T14:08:55.285Z [container-agent] 2025-05-28 14:08:55 ERROR juju-log Failed to execute mysql-shell command
2025-05-28T14:08:55.285Z [container-agent] Traceback (most recent call last):
[...]

PR #639, opened 2025-08-01
[DPE-7421] Fix MySQL legacy interface relation

sinclert-canonical:
This PR adds the  `mysql-root` legacy interface to the tests, in order to make https://github.com/canonical/mysql-k8s-operator/issues/614 visible.

The idea is to relate the MySQL K8s and MySQL test app charms via the legacy `mysql-root` interface **before** the charms finish their deployment. Something that we missed in the MySQL K8s tests (see [example](https://github.com/canonical/mysql-k8s-operator/blob/84deb1fb0cf69eaf1a5a87ba4907b6b2178bd255/tests/integration/relations/test_mysql_root.py#L165-L170)), which only relate applications through the legacy interface **after** the involved charms have been properly deployed.

### Notes

Given that the original `relations/test_mysql_root.py` [module](https://github.com/canonical/mysql-k8s-operator/blob/4f198a8a2787a25da89c1240b3bf7efd1a7d1228/tests/integration/relations/test_mysql_root.py) was dedicated to test **the whole OSM integration**, I moved it to the folder where the COS integration tests are, in order to make space for the simple legacy interface test.

See the test copy + paste effort in commit https://github.com/canonical/mysql-k8s-operator/pull/639/commits/e3202b662378e8790aa02d7c9962074260feb757

### DIFF
--- a/src/relations/mysql.py
+++ b/src/relations/mysql.py
@@ -212,7 +212,7 @@ class MySQLRelation(Object):
         if (
             not self.charm._is_peer_data_set
             or not self.charm.unit_initialized()
-            or self.charm.unit_peer_data.get("member-state") != "online"
+            or not self.charm.unit_peer_data.get("member-state") == "online"
         ):
             logger.info("Unit not ready to execute `mysql` relation created. Deferring")
             event.defer()
--- a/src/relations/mysql_root.py
+++ b/src/relations/mysql_root.py
@@ -8,7 +8,11 @@ import logging
 import typing
 
 from charms.mysql.v0.mysql import MySQLCheckUserExistenceError, MySQLDeleteUsersForUnitError
-from ops.charm import RelationBrokenEvent, RelationCreatedEvent
+from ops.charm import (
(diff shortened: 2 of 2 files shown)


sinclert-canonical:
> OSM is indeed [unsuported](https://charmhub.io/osm), but I am waiting confirmation we can stop testing on it.

Ok, if we get an answer relatively soon, we can use this PR to drop those tests completely (instead of moving them around, which adds noise to the displayed diff).

paulomach:
> > OSM is indeed [unsuported](https://charmhub.io/osm), but I am waiting confirmation we can stop testing on it.
> 
> Ok, if we get an answer relatively soon, we can use this PR to drop those tests completely (instead of moving them around, which adds noise to the displayed diff).

That's the plan

paulomach (approved):

sinclert-canonical:
Clarified with Paulo offline: we need to keep the OSM integration tests for now, at least for every push to the `edge` track.

PR #653, opened 2025-09-03
[MISC] Cleanup MySQL shell timeout workaround

sinclert-canonical:
This PR cleanups up an old workaround for the MySQL shell, where apparently, timeouts were not being applied correctly.

The [linked issue](https://github.com/canonical/operator/issues/1329) describes a misbehavior, but no evidence is provided to backup the claim. In addition, the workaround is not present on the MySQL CLI helper, which calls the same underlying method (see [code](https://github.com/canonical/mysql-k8s-operator/blob/6b91fef5f0915be0b768801a76d370af1bd04579/src/mysql_k8s_helpers.py#L763-L768)). Most probably the misbehavior was due to the version of Juju used at that time, or some other factor.

### DIFF
--- a/src/mysql_k8s_helpers.py
+++ b/src/mysql_k8s_helpers.py
@@ -695,7 +695,6 @@ class MySQL(MySQLBase):
         Returns:
             stdout of the script
         """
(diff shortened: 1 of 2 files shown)


reneradoi (approved):
No concerns.

PR #656, opened 2025-09-10
[DPE-7323] Define MySQL roles page

sinclert-canonical:
This PR defines a new roles documentation page, after the support for Predefined roles. The page has been structured mirroring PostgreSQL roles [page](https://canonical-charmed-postgresql.readthedocs-hosted.com/16/explanation/roles/), but for a couple of changes:

- Charmed MySQL K8s does not create roles depending on whether or not MySQL Router is related.
- Charmed MySQL K8s does not support LDAP authentication.

### DIFF
--- a/docs/explanation/roles.md
+++ b/docs/explanation/roles.md
@@ -13,8 +13,9 @@ There are several definitions of roles in Charmed MySQL K8s:
 ```
 
 ## MySQL K8s roles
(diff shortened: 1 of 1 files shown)


sinclert-canonical:
As privately discussed with @a-velasco, the URL linting script is not able to resolve _some_ of the redirections. This seems to be a bug on documentation side, and they are working on it.

PR #660, opened 2025-09-22
[DPE-8316] Async-replication Jubilant tests

sinclert-canonical:
This PR defines two new sets of Jubilant (JuJu 3 exclusive) tests:

- Async-replication _vanila_ tests, porting existing ones from [this file](https://github.com/canonical/mysql-k8s-operator/blob/main/tests/integration/high_availability/test_async_replication.py).
- Async-replication _upgrade_ tests, new ones within the project.

---

The `new` suffix added to the file name containing the ported integration tests has been added to improve diff visibility on GitHub. Will be removed right before merging.


tests/integration/high_availability/high_availability_helpers_new.py
```diff
@@ -0,0 +1,251 @@
...
+            stop=stop_after_delay(5 * MINUTE_SECS),
+            wait=wait_fixed(10),
+        ):
+            with attempt:
+                unit_max_value = get_mysql_max_written_value(juju, app_name, unit_name)
+                assert unit_max_value > app_max_value, "Writes not incrementing"
+                app_max_value = unit_max_value
+
+
+def exec_k8s_container_command(
```
paulomach, line 49: Does jubilant allow us the same `Juju.exec` or it's machine only?

paulomach (approved):

tests/integration/high_availability/high_availability_helpers_new.py
sinclert-canonical: AFAIK, Jubilant API allow us to execute (`exec` function) either in a Juju unit (i.e. mysql-k8s/0) or in an LXD machine. There is no syntax allowing us to execute a command into a pod **container**, which is what we need in this case.

PR #666, opened 2025-10-08
fix: enabling trust for deployments

deusebio:
## Issue

If there is no `trust`, the mysql deployment hangs. 

## Solution

Just adding the `trust=true` flag in the tf module

## Checklist
- [x] I have added or updated any relevant documentation.
- [x] I have cleaned any remaining cloud resources from my accounts.

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -1,6 +1,7 @@
 resource "juju_application" "mysql_server" {
   name  = var.app_name
   model = var.model_name
+  trust = true
 
   charm {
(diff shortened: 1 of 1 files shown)


deusebio:
Rather critical. KF deployments are broken at the moment. The last modification has introduced some non-backward compatible changes, that we have already mapped (see https://github.com/canonical/charmed-kubeflow-solutions/pull/69) but the deployments are not working because of this. 

Is there any reason why this change is not a beneficial one, and an obvious bug?

PR #680, opened 2025-11-11
[MISC] Adopt packaging standards

astrojuanlu:
PEP 621 ([project] table) and PEP 735 (dependency groups). Requires newer Poetry version.

Notice that this is similar to #613, but it's much less disruptive because it doesn't change our workflow tool (still Poetry). As such, everything should continue working as before. Still, the change is significant because it helps adopt other workflow tools locally.

## Checklist
- [ ] I have added or updated any relevant documentation.
- [ ] I have cleaned any remaining cloud resources from my accounts.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,10 +6,10 @@ name = "mysql-k8s-operator"  # Charm is not meant to be packaged
 dependencies = [
   "ops~=2.15",
   "lightkube~=0.15.0",
-  "tenacity>=8.2.2,<9",
-  "boto3>=1.28.22,<2",
-  "jinja2>=3.1.2,<4",
-  "pyyaml>=6.0.2,<7",
+  "tenacity~=8.2",
+  "boto3~=1.28",
+  "jinja2~=3.1",
+  "pyyaml~=6.0",
 ]
 requires-python = "~=3.10"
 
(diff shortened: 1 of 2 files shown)


astrojuanlu:
Addressed some feedback from @carlcsaposs-canonical 👍🏼 (thank you!)

pyproject.toml
```diff
@@ -1,74 +1,77 @@
 # Copyright 2021 Canonical Ltd.
 # See LICENSE file for licensing details.
 
+[project]
+name = "unused"  # Charm is not meant to be packaged
```
sinclert-canonical: Nit: I would rather have a project name, even if it does not get packaged.

pyproject.toml
```diff
@@ -1,74 +1,77 @@
...
+dependencies = [
+  "ops>=2.15.0,<3",
+  "lightkube>=0.15.0,<0.16",
+  "tenacity>=8.2.2,<9",
+  "boto3>=1.28.22,<2",
+  "jinja2>=3.1.2,<4",
+  "pyyaml>=6.0.2,<7",
```
sinclert-canonical: I see the usage of `>=...,<...` is a direct translation of Poetry caret specifier ([ref](https://python-poetry.org/docs/dependency-specification/#caret-requirements)), however, we are also using Python's compatible specified (`~=`) in the charm-libs group just below. IMHO, the compatible specifier is less verbose, and can be used to express the same version requirements as the `>=...,<...` tuple.

pyproject.toml
```diff
@@ -1,74 +1,77 @@
...
+    # tls_certificates_interface/v1/tls_certificates.py
+    # tls_certificates lib uses a feature only available in cryptography >=42.0.5
+    "cryptography>=42.0.5",
+    "jsonschema",
+    # loki_k8s/v0/loki_push_api.py and prometheus_k8s/v0/prometheus_scrape.py
+    "cosl>=0.0.50",
+    # tempo_coordinator_k8s/v0/charm_tracing.py
+    "opentelemetry-exporter-otlp-proto-http==1.21.0",
+]
+format = ["ruff>=0.12.7,<0.13"]
```
sinclert-canonical: Nit: I would rather have this list expanded, like all the rest within the dependency groups, through the usage of trailing commas.

pyproject.toml
astrojuanlu: It's not possible to express some of our constraints with compatible release specifiers. Take for example the `tenacity` line:

https://github.com/canonical/mysql-k8s-operator/blob/3043204e793b1e14299a6f59055d82d71d769223/pyproject.toml#L9

```
>=8.2.2,<9  # >=8.2.2,==8.*
```

The options are:

```
~=8.2.2     # >=8.2.2,==8.2.*
~=8.2       # >=8.2,  ==8.*
```

so, it either tightens the upper bound too much or loosens the lower bound too much.

sinclert-canonical (approved):

PR #699, opened 2025-12-05
[DPE-8994] Use unit status and ignore app status for data-integrator

astrojuanlu:
There was a difference in behavior between Juju 2.9 and Juju 3.6 that made these tests consistently fail on the former. Units are blocked in both versions, but on Juju 2.9 app appears as "waiting".

As suggested by @paulomach , we're checking only the unit status in those cases.

## Checklist
- [ ] I have added or updated any relevant documentation.
- [ ] I have cleaned any remaining cloud resources from my accounts.


paulomach (approved):
Noice

